PR #2, opened 2019-03-13
Rename projects to remove the word client

MarkZuber:


MarkZuber:
> Sorry, I was not clear. I think that keeping Client is fine for the desktop one, @MarkZuber
> but for not for the Web.
> 
> Or do you want both to have the same name?

Personally, I'd rather keep them consistent and then just not have client.  Also, for ADALv5 file cache, we'd need a third assembly since we don't want to force an ADAL dependency to use the extensions.  so do we want a Microsoft.Identity.AdalExtensions?

jmprieur:
> so do we want a Microsoft.Identity.AdalExtensions?

Or Microsoft.Identity.Extensions.Adal? or Microsoft.Identity.Extensions.Legacy ?

PR #7, opened 2019-03-19
Initial commit of cache management

JPWilli:
This is a port from the Visual Studio IdentityService. Please review with an eye for any holdout VS specifics.


src/Microsoft.Identity.Extensions.Msal/MsalCacheStorage.cs
```diff
@@ -0,0 +1,474 @@
...
+        /// Initializes a new instance of the <see cref="MsalCacheStorage"/> class.
+        /// </summary>
+        /// <param name="creationProperties">Properties for creating the cache storage on disk</param>
+        /// <param name="cacheFilePath">cache file path for msal cache storage</param>
+        /// <param name="instanceName">instance name for msal cache storage</param>
+        /// <param name="logger">logger</param>
+        public MsalCacheStorage(MsalStorageCreationProperties creationProperties, string cacheFilePath, string instanceName, TraceSource logger)
+        {
+            _creationProperties = creationProperties;
+            this._logger = logger ?? throw new ArgumentNullException(nameof(logger));
```
MarkZuber: please avoid "this."

MarkZuber:
> // Copyright (c) Microsoft. All rights reserved.

Everywhere:  our open source header needs to be:

// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.


---
Refers to: src/Microsoft.Identity.Extensions.Msal/MsalCacheStorage.cs:1 in d67785d. [](commit_id = d67785dce8d192bc4f0bd5f44ae9dbdade1c8953, deletion_comment = False)

src/Microsoft.Identity.Extensions.Msal/MsalCacheStorage.cs
```diff
@@ -0,0 +1,474 @@
...
+        private IntPtr _libsecretSchema = IntPtr.Zero;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MsalCacheStorage"/> class.
+        /// </summary>
+        /// <param name="creationProperties">Properties for creating the cache storage on disk</param>
+        /// <param name="cacheFilePath">cache file path for msal cache storage</param>
+        /// <param name="instanceName">instance name for msal cache storage</param>
+        /// <param name="logger">logger</param>
+        public MsalCacheStorage(MsalStorageCreationProperties creationProperties, string cacheFilePath, string instanceName, TraceSource logger)
```
MarkZuber: So, the instance name is only used if cacheFilePath is not specified?  that means that cacheFilePath and instanceName are mutually exclusive.  Could we improve the signature of this API to move one or both of these into the MsalStorageCreationProperties?  Then the PropertyBuilder could validate that TraceSource is actually sent in, and that only one or the other of cacheFilePath or instanceName is specified.

src/Microsoft.Identity.Extensions.Msal/MsalCacheStorage.cs
```diff
@@ -0,0 +1,474 @@
...
+                logger.TraceEvent(TraceEventType.Information, /*id*/ 0, FormattableString.Invariant($"Problem getting last file write time for missing file, trying temp path('{Path.GetTempPath()}'). {e.Message}"));
+                _fileNotFoundOffset = File.GetLastWriteTimeUtc(Path.Combine(Path.GetTempPath(), FormattableString.Invariant($"{Guid.NewGuid().FormatGuidAsString()}.dll")));
+            }
+
+            if (string.IsNullOrWhiteSpace(cacheFilePath))
+            {
+                try
+                {
+                    string localDataPath = SharedUtilities.GetDefaultArtifactPath();
+                    logger.TraceEvent(TraceEventType.Information, /*id*/ 0, FormattableString.Invariant($"Using the home directory '{localDataPath}'"));
```
MarkZuber: nit: use "root" instead of "home" for directory since GetDefaultArtifactPath may or may not return a path under the user's home directory and "home directory" has specific meaning.

src/Microsoft.Identity.Extensions.Msal/MsalCacheStorage.cs
```diff
@@ -0,0 +1,474 @@
...
+            {
+                if (!alreadyLoggedException)
+                {
+                    _logger.TraceEvent(TraceEventType.Error, /*id*/ 0, FormattableString.Invariant($"An exception was encountered while reading data from the {nameof(MsalCacheStorage)} : {e}"));
+                }
+
+                ClearCore();
+            }
+
+            // If the file does not exist this
```
MarkZuber: incomplete comment.

src/Microsoft.Identity.Extensions.Msal/MsalCacheStorage.cs
```diff
@@ -0,0 +1,474 @@
...
+
+        private void ClearCore()
+        {
+            _logger.TraceEvent(TraceEventType.Information, /*id*/ 0, "Clearing cache");
+            bool cacheFileExists = File.Exists(CacheFilePath);
+            _logger.TraceEvent(TraceEventType.Information, /*id*/ 0, FormattableString.Invariant($"ReadDataCore Cache file exists '{cacheFileExists}'"));
+
+            TryProcessFile(() =>
+            {
+                    _logger.TraceEvent(TraceEventType.Information, /*id*/ 0, "Before deleting the cache file");
```
MarkZuber: nit: this is indented too deep.

src/Microsoft.Identity.Extensions.Msal/MsalCacheStorage.cs
```diff
@@ -0,0 +1,474 @@
...
+                    try
+                    {
+                        File.Delete(CacheFilePath);
+                    }
+                    catch (Exception e)
```
MarkZuber: Do you really want to catch all Exception types here?

src/Microsoft.Identity.Extensions.Msal/MsalCacheStorage.cs
```diff
@@ -0,0 +1,474 @@
...
+        private IntPtr GetLibsecretSchema()
+        {
+            if (_libsecretSchema == IntPtr.Zero)
+            {
+                _logger.TraceEvent(TraceEventType.Information, /*id*/ 0, "Before creating libsecret schema");
+
+                _libsecretSchema = Libsecret.secret_schema_new(
+                    name: _creationProperties.KeyringSchemaName,
+                    flags: (int)Libsecret.SecretSchemaFlags.SECRET_SCHEMA_DONT_MATCH_NAME,
+                    attribute1: "string1",
```
MarkZuber: is "string1" and "string2" the data you want here?  Worth a comment as to why these values are used and whether they matter.

src/Microsoft.Identity.Extensions.Msal/MsalStorageCreationProperties.cs
```diff
@@ -0,0 +1,110 @@
...
+        public readonly string MacKeyChainServiceName; //= "Microsoft.Developer.IdentityService";
+        public readonly string MacKeyChainAccountName; //= "MSALCache";
+        public readonly string KeyringSchemaName; //= "msal.cache";
+        public readonly string KeyringCollection; //= "default";
+        public readonly string KeyringSecretLabel; //= "MSALCache";
+        public readonly string KeyringAttribute1; //= "Microsoft.Developer.IdentityService";
+        public readonly string KeyringAttribute2; //= "1.0.0.0";
+    }
+
+    public class MsalStorageCreationPropertiesBuilder
```
MarkZuber: Prefer to keep one class per file

src/Microsoft.Identity.Extensions.Msal/MsalStorageCreationProperties.cs
```diff
@@ -0,0 +1,110 @@
...
+            string macKeyChainServiceName,
+            string macKeyChainAccountName,
+            string keyringSchemaName,
+            string keyringCollection,
+            string keyringSecretLabel,
+            string keyringAttribute1,
+            string keyringAttribute2
+            )
+        {
+            MSALCacheFileName = msalCacheFileName;
```
MarkZuber: Nit: consistency.  We call it Msal everywhere else, can we keep them all as Msal for .NET typing guidelines?

src/Shared/MacKeyChain.cs
```diff
@@ -0,0 +1,232 @@
...
+                                                                          accountNameLength: (uint)accountName.Length,
+                                                                          accountName: accountName,
+                                                                          passwordLength: out valueLength,
+                                                                          passwordData: out valuePtr,
+                                                                          itemRef: out itemRef);
+
+                if (status != NativeMethods.errSecSuccess
+                    && status != NativeMethods.errSecItemNotFound)
+                {
+                    throw new Exception(string.Format(CultureInfo.CurrentCulture, Resources.MacKeyChainFindFailed, status));
```
MarkZuber: Why are you throwing Exception?  That's usually bad practice.  Should we create an exception type that developers can catch so they don't have to catch the base Exception classs?

src/Shared/MacKeyChain.cs
```diff
@@ -0,0 +1,232 @@
...
+        {
+            IntPtr itemRef = IntPtr.Zero;
+            IntPtr valuePtr = IntPtr.Zero;
+
+            try
+            {
+                // check if the key already exists
+                uint valueLength = 0;
+                int status = NativeMethods.SecKeychainFindGenericPassword(
+                                                                          keychainOrArray: IntPtr.Zero,
```
MarkZuber: nit: strange indenting here and on other methods.

src/Shared/NativeMethods.cs
```diff
@@ -0,0 +1,124 @@
...
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Identity.Extensions
+{
+    /// <summary>
+    /// Helper native methods for MAC keychain
+    /// </summary>
+    internal static class NativeMethods
```
MarkZuber: Should this be MacNativeMethods since you also have LinuxNativeMethods?

src/Shared/SharedUtilities.cs
```diff
@@ -0,0 +1,337 @@
...
+{
+    /// <summary>
+    /// A set of utilities shared between service and client
+    /// </summary>
+    internal static class SharedUtilities
+    {
+        /// <summary>
+        /// default base cache path
+        /// </summary>
+        private const string DefaultBaseCachePath = @".IdentityService";
```
MarkZuber: nit:  don't need the @ prefix

src/Shared/SharedUtilities.cs
```diff
@@ -0,0 +1,337 @@
...
+        private static readonly string s_userEnvVar = Environment.GetEnvironmentVariable("USER");
+        private static readonly string s_lNameEnvVar = Environment.GetEnvironmentVariable("LNAME");
+        private static readonly string s_usernameEnvVar = Environment.GetEnvironmentVariable("USERNAME");
+
+        /// <summary>
+        /// For the case where we want to log an exception but not handle it in a when clause
+        /// </summary>
+        /// <param name="loggingAction">Logging action</param>
+        /// <returns>false always in order to skip the exception filter</returns>
+        public static bool LogExceptionAndDoNotHandler(Action loggingAction)
```
MarkZuber: nit:  this naming is a bit weird.  Should this be LogExceptionAndDoNothingHandler() or LogExceptionAndDoNotHandle() ?

src/Shared/SharedUtilities.cs
```diff
@@ -0,0 +1,337 @@
...
+        /// <summary>
+        /// Is this a MAC platform
+        /// </summary>
+        /// <returns>A value indicating if we are running on mac or not</returns>
+        public static bool IsMacPlatform()
+        {
+#if NET46
+            // we have to also check for PlatformID.Unix because Mono can sometimes return Unix as the platform on a Mac machine.
+            // see http://www.mono-project.com/docs/faq/technical/
+            return Environment.OSVersion.Platform == PlatformID.MacOSX || Environment.OSVersion.Platform == PlatformID.Unix;
```
MarkZuber: So is it possible then that this would trigger on a Unix platform and so code there would try to use the keychain interop?

src/Shared/SharedUtilities.cs
```diff
@@ -0,0 +1,337 @@
...
+        }
+
+        /// <summary>
+        /// Gets the users home directory
+        /// </summary>
+        /// <returns>Home directory</returns>
+        public static string GetIdentityServiceRootDirectory()
+        {
+            bool isWindows = Environment.OSVersion.Platform == PlatformID.Win32NT;
+            return GetIdentityServiceUserRootDirectory(isWindows);
```
MarkZuber: Shouldn't this call IsWindowsPlatform() instead of checking the PlatformID again?

src/Shared/SharedUtilities.cs
```diff
@@ -0,0 +1,337 @@
...
+        }
+
+        /// <summary>
+        /// Generate the default file location
+        /// </summary>
+        /// <param name="windowsOS">Is the operating system windows or not</param>
+        /// <returns>Home directory</returns>
+        internal static string GetIdentityServiceUserRootDirectory(bool windowsOS)
+        {
+            return !windowsOS ? SharedUtilities.GetUserHomeDirOnUnix() : Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
```
MarkZuber: or better yet, have this method not take a boolean but instead just call IsWindowsPlatform() within it?

src/Shared/SharedUtilities.cs
```diff
@@ -0,0 +1,337 @@
...
+        {
+            return Path.Combine(SharedUtilities.GetUserHomeDirOnUnix(), ".ServiceHub");
+        }
+
+        /// <summary>
+        /// Determines if the authority is an ADFS authority
+        /// </summary>
+        /// <param name="authority">Uri to check if it is ADFS or not</param>
+        /// <returns>true if the authority is an adfs authority, false otherwise</returns>
+        internal static bool IsADFSAuthority(Uri authority)
```
MarkZuber: Ideally this is called from MSAL instead of re-implementing this type of logic here.  Can you please add a TODO: to fix this once MSAL has the InternalsVisibleTo set?

MsalNetExt.sln
```diff
@@ -1,25 +1,35 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.26124.0
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.28711.60
 MinimumVisualStudioVersion = 15.0.26124.0
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{FF6F9ABE-66F6-42F0-BB6C-D0C8C7372150}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Identity.Extensions", "src\Microsoft.Identity.Extensions\Microsoft.Identity.Extensions.csproj", "{3AEFAFAC-CF36-4FF9-9A64-0F4FCEB609BE}"
```
MarkZuber: Since you've renamed the assemblies, we'll need to update MSAL and ADAL bits to add InternalsVisibleTo to these assemblies instead of the ones we already had.

tests/Microsoft.Identity.Extensions.Msal.UnitTests/MsalCacheStorageTests.cs
```diff
@@ -0,0 +1,172 @@
...
+    public class MsalCacheStorageTests
+    {
+        public static readonly string CacheFilePath = Path.Combine(Path.GetTempPath(), Path.GetTempFileName());
+        private readonly TraceSource _logger = new TraceSource("TestSource");
+        private static MsalStorageCreationProperties s_storageCreationProperties;
+
+        [ClassInitialize]
+        public static void ClassInitialize(TestContext _)
+        {
+            var builder = new MsalStorageCreationPropertiesBuilder("msal.cache");
```
MarkZuber: nit:  why not use the builder as a builder:

s_storageCreationProperties = new MsalStorageCreationPropertiesBuilder("msal.cache").WithMacKeyChainAccountName(...).with(...).Build();

tests/Microsoft.Identity.Extensions.Msal.UnitTests/MsalCacheStorageTests.cs
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        [TestCleanup]
+        public void TestCleanup()
+        {
+            CleanTestData();
+        }
+
+        [TestMethod]
+        public void MsalNewStoreNoFile()
```
MarkZuber: nit:  i don't think you need Msal as a prefix on each method since this is in the MsalCacheStorageTests already so we know it's Msal.

src/Microsoft.Identity.Extensions.Adal/AdalStorageCreationProperties.cs
```diff
@@ -0,0 +1,110 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Identity.Extensions.Adal
+{
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
```
MarkZuber: should add comments instead of disabling the warning.

tests/Microsoft.Identity.Extensions.Adal.UnitTests/AdalCacheStorageTests.cs
```diff
@@ -0,0 +1,172 @@
...
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Identity.Extensions.Adal.UnitTests
+{
+    [SuppressMessage("StyleCop.CSharp.DocumentationRules", "SA1600:ElementsMustBeDocumented", Justification = "Unit Testing")]
```
MarkZuber: ideally, shouldn't need to do this.  The Directory.Build.props should have it set so that documentation rules aren't required for public code in the tests/ subdir.  Can you please verify?

MsalNetExt.sln
MarkZuber: I have pull requests in msal/adal out for this now.  so i'll mark it resolved in this PR.  wanted to capture it so i didn't forget. 

---
In reply to: [267142506](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/pull/7#discussion_r267142506) [](ancestors = 267142506)

src/Microsoft.Identity.Extensions.Msal/MsalCacheStorage.cs
JPWilli: In this case, in Unix systems it does actually call into SharedUtilities.GetUserHomeDirOnUnix(), so I feel home is somewhat warranted. I do admit that the other alternative (%localappdata%) is a bit more distant from the concept of 'home'. Thoughts?

src/Microsoft.Identity.Extensions.Msal/MsalCacheStorage.cs
JPWilli: There are 7 kinds of exceptions that can get thrown here, and they're not all nicely derived from a parent. Also we're simply logging and moving on, so unless there's a compelling reason to be more specific I'd prefer to leave it as is.

src/Shared/SharedUtilities.cs
JPWilli: Going with the shorter one

MsalNetExt.sln
JPWilli: Can I ask here for a quick review of the names I went to? I went this way to make the shared project namespace decisions easier, but I'm open to changing.

tests/Microsoft.Identity.Extensions.Msal.UnitTests/MsalCacheStorageTests.cs
JPWilli: In this case, since this code owns the builder, having the intermediate steps cached out to the local variable every line was intentional so that we could debug the builder code too if needed.

tests/Microsoft.Identity.Extensions.Msal.UnitTests/MsalCacheStorageTests.cs
JPWilli: I was having trouble keeping msal and adal tests clear in the test explorer, since they had the same names, so I renamed to be able to see at a glance. That said, the test explorer does have group-by-project which would also solve the same problem. Any strong opinions knowing my initial intent?

MsalNetExt.sln
MarkZuber: I'm totally fine with the updated names and we've updated the InternalsVisibleTo on our end to work with these new names (will be available in our msal.net 3.0.1 release we're validating now).

---
In reply to: [267162655](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/pull/7#discussion_r267162655) [](ancestors = 267162655)

src/Shared/SharedUtilities.cs
JPWilli: Unused code, I'll just delete

src/Shared/SharedUtilities.cs
JPWilli: That is possible, but I haven't been able to see a better answer available in .netcore.

src/Shared/MacKeyChain.cs
JPWilli: Here, we really don't know much more than "an exception was thrown". Also, adding a public type in the shared code is not a good idea. Not acting on this now.

src/Microsoft.Identity.Extensions.Msal/MsalCacheStorage.cs
JPWilli: Made into a KeyValuePair on the properties object

src/Microsoft.Identity.Extensions.Msal/MsalCacheStorage.cs
JPWilli: Added a directory field to the builder and removed the CacheFilePath entirely

PR #9, opened 2019-03-26
Jpwilli/locking

JPWilli:
Adds locking to cache operations
Reverts projects to .net4.5
Upgrades projects to netcoreapp2.1 (redundant with Mark's other PR)
Adds tests


MarkZuber:
> using System;

please add license blurb

---
Refers to: src/Shared/CrossPlatLock.cs:1 in d756837. [](commit_id = d756837b904e3e1c621279e3ffc719a5f2b5fce5, deletion_comment = False)

MarkZuber:
> using System;

license

---
Refers to: tests/Microsoft.Identity.Extensions.Msal.UnitTests/MockTokenCache.cs:1 in d756837. [](commit_id = d756837b904e3e1c621279e3ffc719a5f2b5fce5, deletion_comment = False)

tests/Microsoft.Identity.Extensions.Msal.UnitTests/MockTokenCache.cs
```diff
@@ -0,0 +1,81 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.Identity.Client;
+using Microsoft.Identity.Client.Cache;
+
+#pragma warning disable CS0618 // Type or member is obsolete (CacheData)
+namespace Microsoft.Identity.Extensions.Msal.UnitTests
+{
+    class MockTokenCache : ITokenCache
```
MarkZuber: please add visibility specifier (private in this case)

src/Shared/CrossPlatLock.cs
```diff
@@ -0,0 +1,98 @@
...
+                    Thread.Sleep(LockfileRetryWait);
+                }
+            }
+
+            if (fileStream == null && exception != null)
+            {
+                throw new InvalidOperationException("Could not get access to the shared lock file.", exception);
+            }
+
+            _lockFileStream = fileStream;
```
MarkZuber: if file stream is null and exception is null, should you ever reach here?  shouldn't above check just be for fileStream being null and you can optionally add the inner exception?

src/Microsoft.Identity.Extensions.Adal/AdalCache.cs
```diff
@@ -66,11 +68,16 @@ private void BeforeAccessNotification(TokenCacheNotificationArgs args)
         {
             _logger.TraceEvent(TraceEventType.Information, /*id*/ 0, $"Before access");
 
+            _logger.TraceEvent(TraceEventType.Information, /*id*/ 0, $"Acquiring lock for token cache");
+            _cacheLock = new CrossPlatLock(
+                Path.GetFileNameWithoutExtension(_store.CreationProperties.CacheFileName),
+                Path.Combine(_store.CreationProperties.CacheDirectory, _store.CreationProperties.CacheFileName) + ".lockfile");
```
MarkZuber: is there a story for what happens when this .lockfile exists but the owning process terminates and this file remains?  How does the system recover in that case.  Should it contain a process id or a last accessed timestamp so that if it's "old" then we can kill the lockfile and proceed?

src/Microsoft.Identity.Extensions.Msal/MsalCacheHelper.cs
```diff
@@ -27,7 +28,8 @@ public class MsalCacheHelper
             return new MsalCacheStorage(s_storageCreationProperties, logger: s_staticLogger.Value);
         });
 
-        private static MsalStorageCreationProperties s_storageCreationProperties;
+        internal static MsalStorageCreationProperties s_storageCreationProperties;
+        internal CrossPlatLock _cacheLock;
```
MarkZuber: if you're going to make these internal so that test code can access it, I'd recommend to make this (and _userTokenCache) into {get;} properties instead of using fields.

src/Microsoft.Identity.Extensions.Adal/AdalCache.cs
JPWilli: The current plan is all participants interested in locking should take an exclusive file lock on that lockfile. If it is not present it should be created. If it exists, but can be opened, that is considered 'unlocked'

src/Microsoft.Identity.Extensions.Msal/MsalCacheHelper.cs
JPWilli: I ended up not needing these for tests. Will revert back to private, but will make _cacheLock a property.

src/Microsoft.Identity.Extensions.Adal/AdalCache.cs
MarkZuber: i misread the intention of this file.  this looks good as the in-memory lock on the file itself will dispose at process termination.

---
In reply to: [269304184](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/pull/9#discussion_r269304184) [](ancestors = 269304184)

PR #13, opened 2019-03-30
Add token provider chains with IConfiguration and ILogger

devigned:
This is a pull request for chained token provider support which provides a simplified path for server to server authentication in Azure, running on servers outside of Azure and on developers' local machine.

### Problem
Developers want to be able to author code once and not change code between local development, testing in the CI environment, and running in production. We need to have an abstraction to make life easier for the developer. 

### Goals
- Simplify developer experience by combining several token acquisition strategies into a single, simple interface.
- Enable developers to write code once and not change it between environments.
- Create a single sign-on like experience for developers such that if they have an ambient identity from Microsoft developer tools they already have authenticated that identity will be used for local development rather than requiring the creation of a service principal.
- Standardize environment variable names so that we can simplify the guidance for injecting configuration as is best practice.

### High level design
A `Chain` consists of a list of `Probes`. A `Probe` can determine if it is available for use (has all data needed to successfully produce a token), and if available, the probe can produce a `TokenProvider` that is able to fetch a `Token` for a list of scopes. A `Token` consists of an `AccessToken` and an `Expiry`. A `Chain` is also a `TokenProvider` and when asked for a token will query each of it's `Probes` to determine if any are available. The first probe that is available will be used to produce a `TokenProvider`, and the provider will be used to return a token to the consumer.

- In Providers:
  - Managed Identity Provider: probe for managed identity service enabling user assigned and default identities on VMs. Also enable probing for environmental, managed identity configuration in App Service.
- Developer's Machine:
  - `SharedTokenCacheProvider` will provide a local developer single sign-on like experience for a collection of developer tools. This is a work in progress and waiting on some features in MSAL to properly implement.
- Anywhere:
  - Service Principal Provider: probe for Service Principal environment variables and configure a token provider to simplify token acquisition.

For a little more background on the goals and problem statement, please see: https://github.com/Azure/azure-auth-wg/blob/master/docs/sdk_auth_helpers.md

### Impersonation local application development (Future?)
When a developer is building applications locally, it would be super useful to be able to express the rights required for an application declarative and enable some form of impersonation to limit the rights the current developer has to only the rights required for the application. If an application developer is unable to replicate the right required locally, they will likely run into issues when deploying the application to be run under a managed identity or service principal with a reduced set of rights, which is often the case since developers generally have more rights than applications.

This feature would require work to define an application manifest or some other way to declare rights, coordination with deployment pipelines and coordination with identity team & libraries.

---

I'm sure there will be a lot of comments. I'm committed to implementing all guidance. Thank you ahead of time for reviewing this PR.

see also: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/924

/cc @jmprieur @cormacpayne @marstr


src/Microsoft.Identity.Client.Extensions.Msal/Providers/Exponential.cs
```diff
@@ -0,0 +1,177 @@
...
+
+        /// <summary>
+        /// Create an instance of an ExponentialBackoff
+        /// </summary>
+        /// <param name="maxRetries">maximum number of retries</param>
+        /// <param name="delayMilliseconds">initial delay in milliseconds</param>
+        /// <param name="maxDelayMilliseconds">maximum delay in milliseconds</param>
+        public ExponentialBackoff(int maxRetries, int delayMilliseconds, int maxDelayMilliseconds)
+        {
+            if(delayMilliseconds <= 0)
```
cormacpayne: @devigned any reason we're doing validation on just this parameter and no others? Also, do we want to do this kind of validation in the `RetryWithExponentialBackoff` constructor as well?

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ManagedIdentityTokenProvider.cs
```diff
@@ -0,0 +1,329 @@
...
+            if(_checkVMListening && !IsVMManagedIdentityListening())
+            {
+                return false;
+            }
+
+            try
+            {
+                // if service is listening on VM IP check if a token can be acquired
+                var provider = BuildInternalProvider(maxRetries: 2, httpClient: _httpClient);
+                var token = await provider.GetTokenAsync(new List<string> { @"https://management.azure.com//.default" }).ConfigureAwait(false);
```
cormacpayne: @devigned since it's a pretty common scope, I think it'd be worth adding `https://management.azure.com/.default` to the `Constants.cs` file

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ServicePrincipalProvider.cs
```diff
@@ -0,0 +1,279 @@
...
+        // Async method lacks 'await' operators and will run synchronously
+        /// <inheritdoc />
+        public Task<bool> AvailableAsync() => Task.FromResult(IsClientSecret() || IsClientCertificate());
+
+
+        /// <inheritdoc />
+        public async Task<IToken> GetTokenAsync(IEnumerable<string> scopes = null)
+        {
+            var provider = await ProviderAsync().ConfigureAwait(false);
+            return await provider.GetTokenAsync().ConfigureAwait(false);
```
cormacpayne: @devigned just want to confirm, do we want to pass `scopes` to the `GetTokenAsync` call here?

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ServicePrincipalProvider.cs
```diff
@@ -0,0 +1,279 @@
...
+                // If the certificate is provided as base64 encoded string in env, decode and hydrate a x509 cert
+                var decoded = Convert.FromBase64String(_config.CertificateBase64);
+                cert = new X509Certificate2(decoded);
+            }
+            else
+            {
+                // Try to use the certificate store
+                var store = new X509Store(StoreNameWithDefault, StoreLocationFromEnv);
+                store.Open(OpenFlags.ReadOnly);
+                var certs = !string.IsNullOrEmpty(_config.CertificateSubjectDistinguishedName) ?
```
cormacpayne: @devigned minor: would we be able to change the next few lines to the following:

```cs
cer = !string.IsNullOrEmpty(_config.CertificateSubjectDistinguishedName) ?
            store.Certificates.Find(X509FindType.FindBySubjectDistinguishedName, _config.CertificateSubjectDistinguishedName, true).FirstOrDefault() :
            store.Certificates.Find(X509FindType.FindByThumbprint, _config.CertificateThumbprint, true).FirstOrDefault();
if (cert == null)
{
    throw new InvalidOperationException(...);
}
```

This will get rid of the need to do `cert = certs[0]` and clean-up the if-statement logic above.

src/Microsoft.Identity.Client.Extensions.Msal/Providers/SharedTokenCacheProvider.cs
```diff
@@ -0,0 +1,83 @@
...
+namespace Microsoft.Identity.Extensions.Msal.Providers
+{
+    /// <summary>
+    /// SharedTokenCacheProbe (wip) provides shared access to tokens from the Microsoft family of products.
+    /// This probe will provided access to tokens from accounts that have been authenticated in other Microsoft products to provide a single sign-on experience.
+    /// </summary>
+    public class SharedTokenCacheProvider : ITokenProvider
+    {
+        private static readonly string CacheFilePath =
+            Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "msal.cache");
```
cormacpayne: @devigned this looks like the correct location for the cache based on [this snippet](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/6d81bbe5da72786445c320a86f22d6e38cc6f53d/src/Shared/SharedUtilities.cs#L113-L118), but I do want to confirm with @MarkZuber and @JPWilli

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ServicePrincipalProvider.cs
devigned: I would, but `store.Certificates.Find` returns a X509Certificate2Collection which doesn't provide the `IEnumerable<T>` interface.

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ServicePrincipalProvider.cs
devigned: Yep. Good catch. Scopes are no longer optional on any of the `GetTokenAsync` methods.

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ManagedIdentityTokenProvider.cs
devigned: Sure. I'll add it.

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ConfigurationProviderExtension.cs
```diff
@@ -0,0 +1,12 @@
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Identity.Extensions.Msal.Providers
+{
+    internal static class ConfigurationProviderExtension
+    {
+        public static string Get(this IConfigurationProvider config, string key)
+        {
+            return config.TryGet(key, out var val) ? val : null;
```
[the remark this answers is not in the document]
devigned: `bool TryGet(string key, out string value);`

I'll change this to a string. If it offended your eyes, it will likely do the same for others.

src/Microsoft.Identity.Client.Extensions.Msal/Providers/Constants.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        /// <summary>
+        /// Managed Identity loopback IP address
+        /// </summary>
+        public static readonly string ManagedIdentityLoopbackAddress = "169.254.169.254";
+
+        /// <summary>
+        /// Managed Identity metadata endpoint
+        /// </summary>
+        public static readonly string ManagedIdentityMetadataEndpoint = "http://" + ManagedIdentityLoopbackAddress + "/metadata";
```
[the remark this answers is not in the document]
devigned: opting for `$"http://{ManagedIdentityLoopbackAddress}/metadata";`

cool?

src/Microsoft.Identity.Client.Extensions.Msal/Providers/Constants.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        /// <summary>
+        /// Managed Identity
+        /// </summary>
+        public static readonly string ManagedIdentityTokenEndpoint = ManagedIdentityMetadataEndpoint + "/identity/oauth2/token";
```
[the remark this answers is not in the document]
devigned: likewise: `$"{ManagedIdentityMetadataEndpoint}/identity/oauth2/token";`

src/Microsoft.Identity.Client.Extensions.Msal/Providers/Exponential.cs
devigned: will fix

devigned:
@MarkZuber please give this a review when you have a bit of time. Thank you!

.deployment
```diff
@@ -0,0 +1,2 @@
+[config]
```
MarkZuber: (learning for me): who reads this file and what is it for?

src/Microsoft.Identity.Client.Extensions.Abstractions/ITokenProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// ITokenProvider describes the interface for fetching an access token
+    /// </summary>
+    public interface ITokenProvider
+    {
+        /// <summary>
+        /// Check if the provider is available for use in the current environment
+        /// </summary>
+        /// <param name="cancel">Cancellation token for early termination of the operation</param>
+        /// <returns>True if a credential provider can be built</returns>
+        Task<bool> AvailableAsync(CancellationToken cancel = default);
```
MarkZuber: naming suggestion:  since this is a bool, have this be a question, so IsAvailableAsync()

src/Microsoft.Identity.Client.Extensions.Msal/Microsoft.Identity.Client.Extensions.Msal.csproj
```diff
@@ -27,22 +27,33 @@
     <IncludeSymbols>true</IncludeSymbols>
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
   </PropertyGroup>
-
   <Import Project="..\Shared\Shared.projitems" Label="Shared" />
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Identity.Client" Version="3.0.8" />
+    <!-- TODO: Change this to a PackageReference after initial publication of Microsoft.Identity.Client.Extensions.Abstractions -->
```
MarkZuber: since this is built in the same repo, you'll want this to be ProjectReference no matter what, correct?  Since we will want these to be built with the same source in the same build.

src/Microsoft.Identity.Client.Extensions.Msal/Microsoft.Identity.Client.Extensions.Msal.csproj
```diff
@@ -27,22 +27,33 @@
...
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetDesktop)'">
+    <Compile Remove="Providers/*.cs" />
     <Reference Include="System.Security" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetCore)'">
     <PackageReference Include="System.Security.Cryptography.ProtectedData" Version="4.5.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="2.2.0" />
```
MarkZuber: suggestion:  for the ones that are the same version, put them in an <ItemGroup> block that isn't conditional so they're included in every version.  And then only the specific ones per platform are in the conditional ItemGroup blocks.

src/Microsoft.Identity.Client.Extensions.Msal/Providers/Exponential.cs
```diff
@@ -0,0 +1,192 @@
...
+        /// <param name="maxDelayMilliseconds">maximum delay in milliseconds</param>
+        public ExponentialBackoff(int maxRetries, int delayMilliseconds, int maxDelayMilliseconds)
+        {
+            const string errMsgGe = "should be greater than or equal 0";
+            const string errMsgG = "should be greater than 0";
+            if (maxRetries < 0)
+                throw new ArgumentOutOfRangeException(nameof(maxRetries), maxRetries, errMsgGe);
+
+            if (delayMilliseconds <= 0)
+                throw new ArgumentOutOfRangeException(nameof(delayMilliseconds), delayMilliseconds, errMsgG);
```
MarkZuber: please use { } even for single line blocks

src/Microsoft.Identity.Client.Extensions.Msal/Providers/Exponential.cs
```diff
@@ -0,0 +1,192 @@
...
+        /// <returns>a task to be delayed</returns>
+        /// <exception cref="TimeoutException">thrown upon exceeding the max number of retries</exception>
+        public async Task DelayAsync()
+        {
+            if (_retries == _maxRetries)
+            {
+                throw new TooManyRetryAttemptsException();
+            }
+
+            var delay = GetDelay(++_retries);
```
MarkZuber: readability:  can you do the ++ on a separate line?  it better shows intent of when you want to increment

src/Microsoft.Identity.Client.Extensions.Msal/Providers/Exponential.cs
```diff
@@ -0,0 +1,192 @@
...
+            var waitTicks = Math.Min(ticks, _maxDelayTicks);
+            return TimeSpan.FromTicks(waitTicks);
+        }
+    }
+
+    /// <inheritdoc />
+    /// <summary>
+    /// TooManyRetryAttemptsException occurs when a retry strategy exceeds the max number of retries
+    /// </summary>
+    public class TooManyRetryAttemptsException : MsalClientException
```
MarkZuber: One class per file, please.

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ManagedIdentityTokenProvider.cs
```diff
@@ -0,0 +1,400 @@
...
+        }
+
+
+        /// <summary>
+        /// IsAppService tells us if we are executing within AppService with Managed Identities enabled
+        /// </summary>
+        /// <returns></returns>
+        private bool IsAppService()
+        {
+            var vars = new List<string> { _config.ManagedIdentitySecret, _config.ManagedIdentityEndpoint };
```
MarkZuber: nit:  please wrap these values onto separate lines

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ManagedIdentityTokenProvider.cs
```diff
@@ -0,0 +1,400 @@
...
+        /// </summary>
+        /// <param name="resourceUri">Resource URI of the protected API requested to access</param>
+        /// <param name="cancel">Cancellation token for the HTTP requests</param>
+        /// <returns>A token with expiration</returns>
+        public async Task<IToken> GetTokenAsync(string resourceUri, CancellationToken cancel)
+        {
+            return await _client.FetchTokenWithRetryAsync(resourceUri, cancel).ConfigureAwait(false);
+        }
+
+        private void Log(Microsoft.Extensions.Logging.LogLevel level, string message, [CallerMemberName] string memberName = "")
```
MarkZuber: please use string.Empty instead of ""

src/Microsoft.Identity.Client.Extensions.Msal/Providers/Exponential.cs
```diff
@@ -0,0 +1,192 @@
...
+using System.Diagnostics;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.Extensions.Msal.Providers
+{
+    /// <summary>
+    /// RetryWithExponentialBackoff runs a task with an exponential backoff
+    /// </summary>
+    internal sealed class RetryWithExponentialBackoff
```
MarkZuber: Please have the file name match the class name.

Also, I know this exists in some other libraries but it may cause us to have an unnecessary dependency to pull it in.  But do you know which assembly already defines this type of behavior?  I know it's a mechanism for azure table storage but i'm not sure where it lives.  I'm fine having our own implementation like this if it brings in too much, but something is telling me this might be in a simple dependency we could re-use.  Just wanted to verify.

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ManagedIdentityTokenProvider.cs
```diff
@@ -0,0 +1,400 @@
...
+        {
+            _logger?.Log(level, $"{nameof(InternalManagedIdentityCredentialProvider)}.{memberName} :: {message}");
+        }
+    }
+
+    internal abstract class ManagedIdentityClient
```
MarkZuber: one class per file please.

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ServicePrincipalProvider.cs
```diff
@@ -0,0 +1,319 @@
...
+        /// <inheritdoc />
+        public async Task<IToken> GetTokenWithResourceUriAsync(string resourceUri, CancellationToken cancel = default)
+        {
+            var provider = await ProviderAsync().ConfigureAwait(false);
+            Log(Microsoft.Extensions.Logging.LogLevel.Information,  "fetching token");
+            var scopes = new List<string>{resourceUri + "/.default"};
+            return await provider.GetTokenAsync(scopes, cancel).ConfigureAwait(false);
+        }
+
+        private async Task<InternalServicePrincipalTokenProvider> ProviderAsync()
```
MarkZuber: Is this a Get?  Can you please make this method name a verb to describe what it's doing with the provider?  It looks like it's creating it.  So perhaps CreateProviderAsync()?

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ServicePrincipalProvider.cs
```diff
@@ -0,0 +1,319 @@
...
+        {
+            var available = await AvailableAsync().ConfigureAwait(false);
+            if (!available)
+            {
+                Log(Microsoft.Extensions.Logging.LogLevel.Information,  "provider is not available");
+                throw new InvalidOperationException("The required environment variables are not available.");
+            }
+
+            var authorityWithTenant = string.Format(CultureInfo.InvariantCulture, AadAuthority.AadCanonicalAuthorityTemplate, _config.Authority, _config.TenantId);
```
MarkZuber: in the past, i've seen bugs with constant strings that have {0} style formats in them.  We solved this by having our string constants files that need these to have the format method in the class itself.  So perhaps you could replace this with AadAuthority.CreateFromAadCanonicalAuthorityTemplate(authority, tenantId) so that the format logic is in one place.

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ServicePrincipalProvider.cs
```diff
@@ -0,0 +1,319 @@
...
+            if (!IsClientCertificate())
+            {
+                Log(Microsoft.Extensions.Logging.LogLevel.Information,  "provider is configured to use client certificates");
+                return new InternalServicePrincipalTokenProvider(authorityWithTenant, _config.TenantId, _config.ClientId, _config.ClientSecret);
+            }
+
+            X509Certificate2 cert;
+            if (!string.IsNullOrWhiteSpace(_config.CertificateBase64))
+            {
+                Log(Microsoft.Extensions.Logging.LogLevel.Information,  $"decoding certificate from base64 directly from environment variable {Constants.AzureCertificateEnvName}");
```
MarkZuber: nit: here and elsewhere, please remove extra space after comma

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ServicePrincipalProvider.cs
```diff
@@ -0,0 +1,319 @@
...
+                else
+                {
+                    Log(Microsoft.Extensions.Logging.LogLevel.Information,  $"finding certificates in store by thumbprint {_config.CertificateThumbprint}");
+                    certs = store.Certificates.Find(X509FindType.FindByThumbprint, _config.CertificateThumbprint, true);
+                }
+
+
+                if (certs.Count < 1)
+                {
+                    var msg = !string.IsNullOrEmpty(_config.CertificateSubjectDistinguishedName)
```
MarkZuber: nit (readability): could you remove the ! and have "var msg = string.IsNullOrEmpty() ? trueCase : falseCase"

src/Microsoft.Identity.Client.Extensions.Abstractions/ITokenProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        /// <summary>
+        /// GetTokenAsync will attempt to fetch a token for a given set of scopes
+        /// </summary>
+        /// <param name="scopes">Scopes requested to access a protected API</param>
+        /// <param name="cancel">Cancellation token for early termination of the operation</param>
+        /// <returns>An access token as a string</returns>
+        Task<IToken> GetTokenAsync(IEnumerable<string> scopes, CancellationToken cancel = default);
```
MarkZuber: So, MSAL has additional parameters that can be sent to AcquireTokenAsync.  ExtraQueryParameters, for example.  How does this interface enable the ability to pass through that kind of content for similar scenarios?  Should we look at having GetTokenAsync and GetTokenWithResourceUriAsync implement the builder pattern for these APIs like MSAL 3.x does so we can provide extensibility of these operations as needed?

tests/WebAppTestWithAzureSDK/.env.sample
```diff
@@ -0,0 +1 @@
+
```
MarkZuber: what is this file for?

tests/WebAppTestWithAzureSDK/.gitignore
```diff
@@ -0,0 +1,11 @@
+azure.json
+
+# Test Infrastructure
+terraform.tfvars
```
MarkZuber: What is terraform and these other files?

tests/WebAppTestWithAzureSDK/Program.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System.IO;
```
MarkZuber: add license comment block

tests/WebAppTestWithAzureSDK/Credentials/DefaultChainServiceCredentials.cs
```diff
@@ -0,0 +1,57 @@
+using System;
```
MarkZuber: license comment block

tests/WebAppTestWithAzureSDK/Credentials/DefaultChainServiceCredentials.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        public DefaultChainServiceCredentials(IConfiguration config = null, ILogger<DefaultChainServiceCredentials> logger = null)
+        {
+            _tokenProvider = new DefaultTokenProviderChain(config: config, logger: logger);
+        }
+
+        public override Task ProcessHttpRequestAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            // TODO: probably should be asking the ServiceClient what the resourceUri is...
+            const string resourceUri = "https://management.azure.com";
```
MarkZuber: You have a comment elsewhere for management.azure.com that the trailing / is important (and it is) but you don't have it here.  is that a bug?

tests/WebAppTestWithAzureSDK/Credentials/DefaultChainServiceCredentials.cs
```diff
@@ -0,0 +1,57 @@
...
+                {
+                    var token = _tokenCache[resourceUri];
+                    if (!token.ExpiresOn.HasValue ||
+                        !(token.ExpiresOn < DateTime.Now.Subtract(new TimeSpan(0, 5, 0))))
+                    {
+                        return _tokenCache[resourceUri];
+                    }
+                }
+
+                var newToken = _tokenProvider.GetTokenWithResourceUriAsync(resourceUri)
```
MarkZuber: You should make this GetTokenAsync and not call GetAwaiter().GetResult().

tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/Providers/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,390 @@
...
+            });
+            var client = new HttpClient(handler);
+            var config = FakeConfiguration(new List<KeyValuePair<string, string>>
+            {
+                new KeyValuePair<string, string>(Constants.ManagedIdentityEndpointEnvName, "http://127.0.0.1/foo"),
+                new KeyValuePair<string, string>(Constants.ManagedIdentitySecretEnvName, "secret")
+
+            });
+            var provider = new ManagedIdentityTokenProvider(httpClient: client, config: config);
+            var token = await provider.GetTokenAsync(new List<string> { "https://management.azure.com/.default" }).ConfigureAwait(false);
```
MarkZuber: you don't have the // for management.azure.com here

.deployment
devigned: This file is used to tell Azure WebApps what to run when the project is `git push`'d up to the repo. Right now, it points to the WebApp that uses the TokenProviderChain to handle Managed Identities.

src/Microsoft.Identity.Client.Extensions.Abstractions/ITokenProvider.cs
devigned: Will do.

src/Microsoft.Identity.Client.Extensions.Msal/Microsoft.Identity.Client.Extensions.Msal.csproj
devigned: Ok. I wasn't quite sure with the abstractions assembly since I'm expecting that would stay relatively static relative to the surrounding projects.

I'll remove the TODO.

src/Microsoft.Identity.Client.Extensions.Msal/Microsoft.Identity.Client.Extensions.Msal.csproj
devigned: These deps are do not support net45, thus were described twice in the core and standard targets. I have consolidated them under a core or standard conditional group.

src/Microsoft.Identity.Client.Extensions.Msal/Providers/Exponential.cs
devigned: Done.

src/Microsoft.Identity.Client.Extensions.Msal/Providers/Exponential.cs
devigned: fixed.

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ManagedIdentityTokenProvider.cs
devigned: done and done.

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ManagedIdentityTokenProvider.cs
devigned: I'm going to use `default` rather than `string.Empty`. In this case, there should be no difference since `var hello = $"hello world!{null}";` concats just fine to `hello world!` and string.Empty is not a compile time constant.

src/Microsoft.Identity.Client.Extensions.Msal/Providers/Exponential.cs
devigned: Will do on the file name / class name.

Per the other libraries... I'm certain there are multiple places where something like this is implemented. I looked around and didn't find a dependency that I though would be innocuous enough to warrant the weight of an additional dependency vs a little bit of code. I definitely defer to you on this. If you have a strong opinion, let me know.

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ManagedIdentityTokenProvider.cs
devigned: heh... figured this was going to come up. You got it!

devigned: Moved each class into it's own file and moved exceptions into their own namespace / directory.

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ServicePrincipalProvider.cs
devigned: Done!

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ServicePrincipalProvider.cs
devigned: oops... Fixed

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ServicePrincipalProvider.cs
devigned: Fixed

src/Microsoft.Identity.Client.Extensions.Abstractions/ITokenProvider.cs
devigned: They could implement the builder pattern, but the interface gets _big_. I was hoping that this abstraction would handle the 90% use case with extreme simplicity.

Would service to service communication scenarios use `ExtraQueryParameters`? If so, what scenarios would they be used?

tests/WebAppTestWithAzureSDK/.env.sample
devigned: I was going to expand on the example and talk more about the deployment script, but decided that would be more than I'd like to get into in this PR. It's already a novel :(.

devigned: Deleted the .env.sample

tests/WebAppTestWithAzureSDK/Program.cs
devigned: Added.

tests/WebAppTestWithAzureSDK/Credentials/DefaultChainServiceCredentials.cs
devigned: Added.

tests/WebAppTestWithAzureSDK/Credentials/DefaultChainServiceCredentials.cs
devigned: It's not a bug, but it would cause the ServiceClientCredentials to only work with Azure Resource Management. While today, this is totally functional with the current code base, this is something the Azure SDK group will need to provide in their ARM HTTP clients.

Basically, I'm using this as an example for what we need to change in Azure SDKs to provide the user experience desired. A proper credential / client interface would need to provide the scopes (or resourceUri in this case) to the ITokenProvider implementation, so consumers don't need to hunt down the scopes or resourceUri for the service client they are using. It's a major pain point for consumers of Azure libraries today.

tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/Providers/ManagedIdentityTests.cs
devigned: Are you referring to the "//" at the end? If so, it still seems to work when I'm using it.

tests/WebAppTestWithAzureSDK/Credentials/DefaultChainServiceCredentials.cs
devigned: It's within a lock block, so no async. Is there another way I should handle that?

tests/WebAppTestWithAzureSDK/.gitignore
devigned: [Terraform](https://www.terraform.io) is the tool that is used to deploy this [file of resources](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/pull/13/files#diff-7befd529ab4b6b6214230b78c7465476). You can [deploy the needed Azure resources as it's specified in the readme](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/pull/13/files#diff-650c1ef00148ad8c608bdfa6ef178bf3R37). It makes it super easy for a developer to get started with all of the resources needed to run a test of the web application. The `.tfvars` and other TF related stuff is the state files that TF keeps around and should not be checked in.

The azure.json file is a configuration file that the AspNetCore app is configured to consume. It could have configuration information that doesn't need to be checked in, thus an ignore.

src/Microsoft.Identity.Client.Extensions.Msal/Providers/Exponential.cs
devigned: Yep, done!

devigned:
@MarkZuber I think I have addressed all of your feedback. If you are satisfied with the answer to a comment, please mark them as resolved (hope that will help keep attention to outstanding issues).

Let me know if there are other things you'd like me to address, and thank you so much for your awesome feedback!

src/Microsoft.Identity.Client.Extensions.Msal/Providers/ServicePrincipalProvider.cs
devigned: Overlooked this one. It is now resolved as requested.

tests/WebAppTestWithAzureSDK/Credentials/DefaultChainServiceCredentials.cs
MarkZuber: We're restructuring our tokencache callbacks to be async and hit the similar issue.  You can see a PR here where we use SemaphoreSlim to construct the locking so that we can still use the proper async pattern.  The GetAwaiter() calls can sometimes cause deadlocks depending on what's going on.

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1088

devigned: Great idea! I've updated the PR with your suggestion. Thank you.

.deployment
brentschmaltz: @devigned is this PR about storing tokens in a cache in a MSI environment?

devigned: @brentschmaltz, it doesn't use the token cache in the managed identity enabled environment.

The progression of TokenProviders is as follows:
1) ServicePrincipalProvider
2) ManagedIdentityProvider
3) SharedTokenCacheProvider

The chain calls 1 and finds no env vars which would satisfy the IsAvailable criteria, then it calls 2 and probes for the manage identity provider. The managed identity provider is found and subsequently used to procure tokens. At no time is the MSAL token cache used since the chain stopped at number 2.

devigned: Also, should I expand more upon the goals in the readme? Were they not clear enough?

brentschmaltz: @devigned I should read them :-)

src/Microsoft.Identity.Client.Extensions.Msal/Providers/DefaultManagedIdentityConfiguration.cs
```diff
@@ -0,0 +1,20 @@
+using Microsoft.Extensions.Configuration;
```
MarkZuber: Add license header here and elsewhere, please.

devigned: I thought I got them all. Must have missed this one. I'll verify each source file.

devigned:
@MarkZuber just rebased and added the license header to the rest of the .cs files. After the build goes green will you be the one merging?

PR #14, opened 2019-04-01
Fewer statics, and clear without public app

JPWilli:
Update MsalCacheHelper to have fewer statics and provide a way to create a helper without registering so that users can clear the cache without a public app.

### DIFF
--- a/src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
+++ b/src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
@@ -91,11 +91,6 @@ namespace Microsoft.Identity.Client.Extensions.Msal
         }
 
         /// <summary>
-        /// Gets whether this instance has had <see cref="RegisterCache(ITokenCache)" /> called.
-        /// </summary>
-        public bool Registerred { get; private set; }
-
-        /// <summary>
         /// Gets a singleton instance of the TokenCacheHelper
         /// </summary>
         /// <param name="tokenCache">Token Cache</param>
@@ -129,7 +124,6 @@ namespace Microsoft.Identity.Client.Extensions.Msal
                     }
                 }
 
-                Registerred = true;
                 _logger.TraceEvent(TraceEventType.Information, /*id*/ 0, $"Done initializing");
             }
         }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
```diff
@@ -18,17 +18,18 @@ public class MsalCacheHelper
         /// </summary>
         private static readonly object s_lockObject = new object();
```
MarkZuber: do you still need the lock to be static?  you've moved everything else out of static here (except the logger) so i'd think you'd want the lock isolated to the MsalCacheHelper instance and not the type.

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
```diff
@@ -109,32 +94,46 @@ public static string UserRootDirectory
...
                 {
-                    throw;
+                    try
+                    {
+                        _logger.TraceEvent(TraceEventType.Information, /*id*/ 0, $"Deserializing data into memory");
+                        _userTokenCache.DeserializeMsalV3(data);
+                    }
+                    catch (Exception e)
+                    {
+                        _logger.TraceEvent(TraceEventType.Information, /*id*/ 0, $"An exception was encountered while deserializing the data during initialization of {nameof(MsalCacheHelper)} : {e}");
```
MarkZuber: shouldn't this be at warning or error level?

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
```diff
@@ -109,32 +94,46 @@ public static string UserRootDirectory
...
+                        _store.Clear();
```
MarkZuber: call Clear() here so that if more logic is added it's picked up / utilized here.

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
JPWilli: I'm not very clear on what kinds of exceptions could come through, and I don't want to make it too chatty. I'll upgrade to warning here at least, so we can differentiate, and if we don't see too many messages there we could consider moving to error later.

MarkZuber: Well, it's "chatty" at informational level anyway.  but this code path is clearing the entire cache store.  it's worth promoting it in the logs as warn/error so that if someone sets the trace filter output to "warning and higher" that they see this is happening.

---
In reply to: [271092807](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/pull/14#discussion_r271092807) [](ancestors = 271092807)

PR #15, opened 2019-04-02
Add target for .NET Standard 2.0 to extension projects

cormacpayne:
Adding a `netstandard2.0` target to both MSAL and ADAL packages to match target used in both client packages.


cormacpayne:
Investigating issue with .NET Standard framework.

PR #16, opened 2019-04-02
Jpwilli/remove identity service path

JPWilli:
Removes the last reference to "IdentityService" from shipping code (although it still exists in tests). Also fixes a small issue with logging where users could give null TraceSource objects and cause problems.


crmann1:
Can you patch and test VS before comitting this so we know its good before publishing another build.

PR #17, opened 2019-04-03
Add claims extensions from identity web samples

MarkZuber:


src/Microsoft.Identity.Client.Extensions.Web/ClaimsPrincipalFactory.cs
```diff
@@ -0,0 +1,47 @@
...
+namespace Microsoft.Identity.Client.Extensions.Web
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public static class ClaimsPrincipalFactory
+    {
+        /// <summary>
+        /// Instantiate a ClaimsPrincipal from an account objectId and tenantId. This can
+        /// we useful when the Web app subscribes to another service on behalf of the user
```
jennyf19: >we [](start = 12, length = 2)

be

src/Microsoft.Identity.Client.Extensions.Web/Microsoft.Identity.Client.Extensions.Web.csproj
```diff
@@ -7,7 +7,7 @@
     <!--This will generate AssemblyVersion, AssemblyFileVersion and AssemblyInformationVersion-->
     <Version>$(ClientSemVer)</Version>
 
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
jmprieur: not 2.2 ?

MarkZuber: Even with 2.2 sdk installed on my machine I was having issues getting VS to build it properly in the tool chain.  I'll look at this again when i get a bit more time and move it to 2.2.  2.2 is where I'd like this to be at as well.

PR #23, opened 2019-04-09
update target frameworks so we can build from other platforms, move t…

MarkZuber:
…o 2.2sdk

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -8,6 +8,17 @@
     <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)/build/SolutionWideAnalyzerConfig.ruleset</CodeAnalysisRuleSet>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+
+    <TargetFrameworkNetDesktop>net45</TargetFrameworkNetDesktop>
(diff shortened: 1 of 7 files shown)


Directory.Build.props
```diff
@@ -1,20 +1,31 @@
...
+    <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
+
+    <LangVersion>7.3</LangVersion>
+    <SignAssembly>true</SignAssembly>
+    <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)/build/MSAL.snk</AssemblyOriginatorKeyFile>
+    <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)/build/SolutionWideAnalyzerConfig.ruleset</CodeAnalysisRuleSet>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+
+    <TargetFrameworkNetDesktop>net45</TargetFrameworkNetDesktop>
```
bgavrilMS: It looks to me that you only have a bit of special code for handling netdesktop (the include system security bit). I'd say you can drop support for netcore then and just use netstandard. Ideally you'd drop support for netdesktop as well and use netstandard...

MarkZuber: We'd do that if we could target netstandard with vs.  this round, we have to have explicit net4.5 support.  once they update their installers, then yes both of these can go to netstandard2.0 and we can drop the other support.

PR #24, opened 2019-04-10
Add event firing when cache file is changed on disk

JPWilli:
Adds a file watcher for the cache file, and fires events detailing which accounts have been added/removed.
Adds encryption to CoreCLR cache file when running in windows.
Other small fixes
Updates to ManualTestApp to facilitate testing


tests/ManualTestApp/Program.cs
```diff
@@ -2,34 +2,38 @@
...
 using Microsoft.Identity.Client;
+using Microsoft.Identity.Client.Extensions.Msal;
 
 namespace ManualTestApp
 {
     class Program
     {
+        private static MsalCacheHelper s_helper;
+
         static void Main(string[] args)
```
MarkZuber: Can you make this an async main so you can do await RunCommandAsync() instead of doing the .Result?

PR #27, opened 2019-05-08
Update version handling code

JPWilli:
There was a bug in how we were using timestamps to track when the cache changed, where they could be inaccurate when queried at high frequencies. Swap out for reading a version file instead.


src/Microsoft.Identity.Client.Extensions.Adal/AdalCacheStorage.cs
```diff
@@ -120,6 +131,12 @@ public byte[] ReadData()
             bool alreadyLoggedException = false;
             try
             {
+                // Guarantee that the version file exists so that we can know if it changes.
+                if (!File.Exists(VersionFilePath))
```
crmann1: If the file does exist I do not see where we read it in.  So this would onlyh work if the file does not exist.

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
```diff
@@ -318,14 +317,14 @@ internal void BeforeAccessNotification(TokenCacheNotificationArgs args)
 
             lock (_lockObject)
             {
-                if ((!_registeredCaches.ContainsKey(args.TokenCache)) || _registeredCaches[args.TokenCache] < _store.LastWriteTime)
+                if ((!_registeredCaches.ContainsKey(args.TokenCache)) || _registeredCaches[args.TokenCache] != _store.LastVersionToken)
```
crmann1: Can you use string.Equals with the ordinal ignore case comparer

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
```diff
@@ -358,9 +357,21 @@ internal void AfterAccessNotification(TokenCacheNotificationArgs args)
                     {
                         _logger.TraceEvent(TraceEventType.Information, /*id*/ 0, $"Before Write Store");
                         byte[] data = args.TokenCache.SerializeMsalV3();
+                        _cachedStoreData = data;
                         _logger.TraceEvent(TraceEventType.Information, /*id*/ 0, $"Serializing '{data.Length}' bytes");
                         _store.WriteData(data);
 
+                        try
+                        {
+                            // Hack: The _store.LastVersionToken is written when we call _store.ReadData, but we know that we're
```
crmann1: Can you not use the work Hack in the code. Just remove it the rest of the comment is fine.

PR #32, opened 2019-06-04
update to msal40 with cache interaction changes

MarkZuber:
Cache serialization with MSAL.NET 4.0 now only happens via TokenNotificationArgs due to locking updates introduced with async token cache callbacks.  So the usage semantics needed to be cleaned up a bit.

### DIFF
--- a/src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
+++ b/src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
@@ -112,10 +112,13 @@ namespace Microsoft.Identity.Client.Extensions.Msal
             if (File.Exists(storageCreationProperties.CacheFilePath))
             {
                 var pca = PublicClientApplicationBuilder.Create(storageCreationProperties.ClientId).Build();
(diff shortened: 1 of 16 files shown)


src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
```diff
@@ -112,10 +112,13 @@ private static async Task<HashSet<string>> GetAccountIdentifiersAsync(StorageCre
...
             {
                 var pca = PublicClientApplicationBuilder.Create(storageCreationProperties.ClientId).Build();
-                var tempCache = new MsalCacheStorage(storageCreationProperties, s_staticLogger.Value);
 
-                // We're using ReadData here so that decryption is gets handled within the store.
-                pca.UserTokenCache.DeserializeMsalV3(tempCache.ReadData());
+                pca.UserTokenCache.SetBeforeAccess((args) =>
+                {
+                    var tempCache = new MsalCacheStorage(storageCreationProperties, s_staticLogger.Value);
+                    // We're using ReadData here so that decryption is gets handled within the store.
```
jennyf19: >is gets [](start = 68, length = 7)

"is gets" ? lol.

PR #33, opened 2019-06-13
fix null argument exception when logger is not provided

devigned:
I ran across an exception when the SharedCacheProvider builds a `MsalCacheStorage` with a null logger. In the case when the logger is null, it seems appropriate to use the static logger rather than throwing an exception.

### DIFF
--- a/tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/Providers/SharedTokenCacheProviderTest.cs
+++ b/tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/Providers/SharedTokenCacheProviderTest.cs
@@ -1,0 +1,41 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Identity.Client.Extensions.Msal.Providers
+{
+    [TestClass]
+    public class SharedTokenCacheProviderTest
+    {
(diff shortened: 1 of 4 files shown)


tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/Providers/SharedTokenCacheProviderTest.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        [TestMethod]
+        [TestCategory("SharedTokenCacheProviderTests")]
+        public async Task ShouldNotBeAvailableWithoutIdentitiesAsync()
+        {
+            var provider = new SharedTokenCacheProvider(_builder);
+            Assert.IsFalse(await provider.IsAvailableAsync().ConfigureAwait(false));
+        }
+
```
MarkZuber: nit: remove extra blank line.

devigned: Thx u and fixed!

devigned:
@MarkZuber since this is a fork PR is the build going to kick off or do I need to do something?

MarkZuber:
> @MarkZuber since this is a fork PR is the build going to kick off or do I need to do something?

Sorry for the delay (I was OOF).  I don't know how AzDO handles forked builds (apparently it doesn't).  I can merge this and see how it works.  Do you not have proper permissions to do in-repo development?

devigned:
Sry, I was OOF yesterday.

I'm pretty sure I have access to push a branch to the repo. I'll just move to doing that for future contributions. I'm just used to running a fork and providing PRs via fork branches.

Thank you for getting this pulled in!

PR #38, opened 2019-07-25
refactor storage and logging to prep for mac cache

MarkZuber:
Currently, MSAL Cache Storage is one file with some "if platform / do logic" for each platform.  There is also the assumption that each platform will have one file / blob.  This may not be true as there may be cases where we split into more than one file or we don't save to a file at all.

So to prep for ensuring we do "TheRightThing(tm)" on each platform and that action is clear, I've made MsalCacheStorage into an abstract class which has a static Create to check the platform and return the appropriate sub-class implementation.

I've also updated the trace logging to add TraceSourceLogger to remove the need to have the "/* id */ 0" call in each of the calls to TraceSource.LogEvent which cleans up readability a bit.

The goal of this PR is that it just restructures/moves code but does not add/remove any functionality from the current execution.  Any actual functional changes will come in follow up pull requests.


src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheStorageLinux.cs
```diff
@@ -0,0 +1,182 @@
...
+using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Identity.Client.Extensions.Msal
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class MsalCacheStorageLinux : MsalCacheStorage
```
henrik-me: private or internal?

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheStorage.cs
```diff
@@ -4,43 +4,68 @@
...
-using System.Security.Cryptography;
 using System.Threading;
 
 namespace Microsoft.Identity.Client.Extensions.Msal
 {
     /// <summary>
     /// Persist cache to file
     /// </summary>
-    public sealed class MsalCacheStorage
+    public abstract class MsalCacheStorage
```
henrik-me: >abstract [](start = 11, length = 8)

What's the reason to use abstract instead of interfaces?

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheStorage.cs
```diff
@@ -4,43 +4,68 @@
...
+            return new TraceSourceLogger((TraceSource)EnvUtils.GetNewTraceSource(nameof(MsalCacheHelper) + "Singleton"));
         });
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MsalCacheStorage"/> class.
+        /// </summary>
+        /// <param name="creationProperties">Properties for creating the cache storage on disk</param>
+        /// <param name="logger">logger</param>
+        /// <returns></returns>
+        public static MsalCacheStorage Create(StorageCreationProperties creationProperties, TraceSource logger = null)
```
henrik-me: asked elsewhere - IMsalStorage ?

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheStorage.cs
```diff
@@ -4,43 +4,68 @@
...
+            {
+                return new MsalCacheStorageMacOs(creationProperties, logger);
+            }
+            else if (SharedUtilities.IsLinuxPlatform())
+            {
+                return new MsalCacheStorageLinux(creationProperties, logger);
+            }
+            else
+            {
+                throw new PlatformNotSupportedException();
```
bgavrilMS: Except for this factory method, it looks like all this class does is decorate functionality with logging. Might be easier to have a separate factory class and a logging decorator rather than using a base class.

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheStorageLinux.cs
```diff
@@ -0,0 +1,182 @@
...
+
+        /// <summary>
+        /// 
+        /// </summary>
+        protected override void ClearCore()
+        {
+            _logger.LogInformation("Clearing cache");
+            DeleteCacheFile(CacheFilePath);
+
+            _logger.LogInformation($"Before deletring secret from linux keyring");
```
bgavrilMS: nit typo

jmprieur:
@henrik-me : do we want to merge this?

henrik-me:
No

bgavrilMS:
@henrik-me - Could you pleasse re-review this?

PR #39, opened 2019-08-06
Proposal to add Microsoft.Identity.Web

jmprieur:


jmprieur:
Closing while we determine the right ship vehicle for the identity.web contents

PR #45, opened 2019-09-17
Removing obsolete code, redirecting to new location

jmprieur:
Following customers feedback (that this folder of the repo was confusing), this PR removes code that has evolved elwhere (and got merged from the changes in this code base)
(See README.md)


jmprieur:
@henrik-me : under src/Microsoft.Identity.Client.Extensions.Web, yes (and adding a README.md)
the alternative would be to copy accross the code from the master sample? 

what would you suggest we do?

PR #49, opened 2019-11-14
Minor cleanup of global.json and test app change

bgavrilMS:
I deleted global.json because: 

- it specified a version of msbuild.sdk.extras, but this project does not use the extras (it's only needed for adding xamarin and uwp targets)
- it specified a version of .net core SDK, which is a somewhat old, but the project builds fine with any SDK

### DIFF
--- a/src/Microsoft.Identity.Client.Extensions.Msal/Microsoft.Identity.Client.Extensions.Msal.csproj
+++ b/src/Microsoft.Identity.Client.Extensions.Msal/Microsoft.Identity.Client.Extensions.Msal.csproj
@@ -35,6 +35,7 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetDesktop)'">
     <Compile Remove="Providers/*.cs" />
+    <None Include="Providers/*.cs" />
     <Reference Include="System.Security" />
   </ItemGroup>
 
--- a/tests/ManualTestApp/Program.cs
+++ b/tests/ManualTestApp/Program.cs
@@ -10,6 +10,9 @@ using Microsoft.Identity.Client.Extensions.Msal;
 
 namespace ManualTestApp
 {
+
+    // Example usage:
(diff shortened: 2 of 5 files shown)


src/Microsoft.Identity.Client.Extensions.Msal/Microsoft.Identity.Client.Extensions.Msal.csproj
```diff
@@ -35,6 +35,7 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetDesktop)'">
     <Compile Remove="Providers/*.cs" />
+    <None Include="Providers/*.cs" />
```
bgavrilMS: This is so that Visual Studio shows all the files. By default Solution Explorer is bound to the first tfm, which is .net45 and on .net45 the extension does not contain all the managed indentity helpers.

tests/ManualTestApp/Program.cs
```diff
@@ -111,7 +114,8 @@ private static async Task LoginWithMsalAsync(string[] args)
                     };
                 }
 
-                var builder = app.AcquireTokenWithDeviceCode(scopes, Utilities.DeviceCodeCallbackAsync);
+                //var builder = app.AcquireTokenWithDeviceCode(scopes, Utilities.DeviceCodeCallbackAsync);
+                var builder = app.AcquireTokenInteractive(scopes);
```
bgavrilMS: It's much faster to test with AcquireTokenInteractive + system browser than with device code flow.

src/Microsoft.Identity.Client.Extensions.Msal/Microsoft.Identity.Client.Extensions.Msal.csproj
```diff
@@ -35,6 +35,7 @@
...
     <Reference Include="System.Security" />
   </ItemGroup>
```
bgavrilMS: Speaking of target frameworks, the extension is built against: 

- net45
- netcore2.1
- netstandard2.0

@crmann1 - There are no differences between netcore and netstandard versions. Is there a reason why we release a dedicated .net core version?

PR #50, opened 2019-11-18
Add a diagram with the cache architecture

bgavrilMS:
Add a diagram showing the 2 main concepts behind the cache architecture: 

- inter process lock using a file lock -> ensures that only 1 process can access the cache store at a time
- actual cache store is: DPAPI encrypted file on Win, KeyChain on Mac, KeyRing on Linux
- a file event watcher is used to be able to generate a CacheChanged event. The CacheChange event reloads the cache and so it is able to figure out which accounts were added and which were deleted
- on Mac and Linux, we don't keep the cache in a file, so a dummy file is used to be able to make it work with the file watcher.


bgavrilMS:
Hopefully this sheds some light on how the extension does caching.

PR #55, opened 2020-01-21
Fix null reference when reading fails

bgavrilMS:
Fix for #57


tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/MsalCacheStorageTests.cs
```diff
@@ -46,6 +48,42 @@ public void TestCleanup()
...
+
+        [TestMethod]
+        public void CacheStorageCanHandleExceptionsWhenReading()
+        {
+            // Arrange
+            var cacheAccessor = NSubstitute.Substitute.For<ICacheAccessor>();
+            var exception = new InvalidOperationException();
+            cacheAccessor.Read().Throws(exception);
+
+            var actualLogger = new TraceSourceLogger(_logger);
```
henrik-me: >actualLogger [](start = 16, length = 12)

while slightly off from a unit test perspective. ... would there be meaning in validating that logger contains exception information as you are asserting that no data is returned? #validatingSupportability

PR #60, opened 2020-02-11
Proposal for Linux KeyRing fallback mechanism.

bgavrilMS:
This is currently blocking Az CLI and is an issue for PowerShell, both of which are CLIs that are expected to work on Linux over SSH. 

@rayluo @SomkaPe @schaabs - please review as the changes would need to be made in your extension libraries as well.

### DIFF
--- a/docs/keyring_fallback_proposal.md
+++ b/docs/keyring_fallback_proposal.md
@@ -77,9 +77,9 @@ Libraries consuming the extension will:
 
 1. create a cache helper with a the normal `KeyRing` setup
 2. call `cacheHelper.VerifyLinuxPersistence()`
-2.1. If this throws an exception, show the user a meaningful message / URL to help page to inform them to secure their secrets storage
-2.2. Create a cache helper using `.WithLinuxPlaintextFile` using a file path that comes from either: 
-a. a well known ENV variable, e.g. DEV_TOOLS_TOKEN_CACHE
-b. if DEV_TOOLS_TOKEN_CACHE is not set, default to a well known location 
+3. If this throws an exception, show the user a meaningful message / URL to help page to inform them to secure their secrets storage
+4. Create a cache helper using `.WithLinuxPlaintextFile` using a file path that comes from either: 
+- a well known env variable, e.g. LINUX_DEV_TOOLS_TOKEN_CACHE
+- if LINUX_DEV_TOOLS_TOKEN_CACHE is not set, default to a well known location 
 
 
--- a/docs/keyring_fallback_proposal.md
+++ b/docs/keyring_fallback_proposal.md
@@ -46,10 +46,10 @@ cacheHelper.RegisterCache(app.UserTokenCache);
 #### Add a method to check persistence on Linux
 
 ```csharp
-void cacheHelper.VerifyLinuxPersistence();
+void cacheHelper.VerifyPersistence();
 ```
 
-This method MUST not affect the token cache. It will attempt to write and read a dummy secret. Different KeyRing attributes will be used so as to not interfere with the real token cache. 
+This method MUST not affect the token cache. It will attempt to write and read a dummy secret. Different storage attributes will be used so as to not interfere with the real token cache (e.g. Windows - different file path, Mac - different account, Linux - differnt keyring attribute)
 
 If this method fails it throws an exception with more details. Typically the failure points are:
 
@@ -69,14 +69,16 @@ If this method fails it throws an exception with more details. Typically the fai
 ```                     
 `Config.CacheFileName` will contain the unprotected cache. 
 
-#### Suggested pattern for extension consumers
+Note: `WithLinuxUnprotectedFile` cannot be used in conjuction with `WithLinuxKeyring` - an exception will be thrown
 
+#### Suggested pattern for extension consumers
+ 
 Libraries consuming the extension will: 
 
 1. create a cache helper with a the normal `KeyRing` setup
-2. call `cacheHelper.VerifyLinuxPersistence()`
+2. call `cacheHelper.VerifyPersistence()`
 3. If this throws an exception, show the user a meaningful message / URL to help page to inform them to secure their secrets storage
-4. Create a cache helper using `.WithLinuxPlaintextFile` using a file path that comes from either: 
+4. Create a cache helper using `.WithLinuxUnprotectedFile` using a file path that comes from either: 
 - a well known env variable, e.g. LINUX_DEV_TOOLS_TOKEN_CACHE
 - if LINUX_DEV_TOOLS_TOKEN_CACHE is not set, default to a well known location 
 
--- a/docs/keyring_fallback_proposal.md
+++ b/docs/keyring_fallback_proposal.md
@@ -21,7 +21,7 @@
         .WithMacKeyChain(
             Config.KeyChainServiceName,
             Config.KeyChainAccountName)
-        .Build|();
+        .Build();
 
 MsalCacheHelper cacheHelper = MsalCacheHelper.Create(storageProperties);
 cacheHelper.RegisterCache(app.UserTokenCache);
@@ -65,10 +65,11 @@ TODO: understand perf impact of this call. If expensive, need to inform consumer
 
 ```csharp
  new StorageCreationPropertiesBuilder(Config.CacheFileName, Config.CacheDir, Config.ClientId) 
-    .WithLinuxPlaintextFile(ConPlaintextFilePath) //new method                     
+    .WithLinuxPlaintextFile(Config.UnprotectedCacheFilePath) //new method                     
     .WithMacKeyChain(...); // no change
                      
 ```                     
+If `UnprotectedCacheFilePath` and  `CacheFileName` are the same, an exception MUST be thrown.
 
 As an alternative, the `.WithLinuxPlaintextFile()` can take 0 arguments, and simply use `Config.CacheFileName`, however this would invalidate Goal 4.
 
--- a/docs/keyring_fallback_proposal.md
+++ b/docs/keyring_fallback_proposal.md
@@ -65,7 +65,7 @@ TODO: understand perf impact of this call. If expensive, need to inform consumer
 
 ```csharp
  new StorageCreationPropertiesBuilder(Config.CacheFileName, Config.CacheDir, Config.ClientId) 
-    .WithLinuxPlaintextFile(Config.UnprotectedCacheFilePath) //new method                     
+    .WithLinuxUnprotectedFile(Config.UnprotectedCacheFilePath) //new method                     
     .WithMacKeyChain(...); // no change
                      
 ```                     
@@ -83,5 +83,3 @@ Libraries consuming the extension will:
 4. Create a cache helper using `.WithLinuxPlaintextFile` using a file path that comes from either: 
 - a well known env variable, e.g. LINUX_DEV_TOOLS_TOKEN_CACHE
 - if LINUX_DEV_TOOLS_TOKEN_CACHE is not set, default to a well known location 
-
-
--- a/docs/keyring_fallback_proposal.md
+++ b/docs/keyring_fallback_proposal.md
@@ -3,7 +3,7 @@
 **Context:** Today, the MSAL extension libraries (.net, java, python, javascript) store secrets in KeyRings via LibSecret. These components not available on all Linux distros and they cannot be started when connected via SSH.
 
 **Proposal:** Extensions are to provide a mechanism for products to detect if this secret storage is usable. If it is not, extensions are to write the token cache to an unecrypted file. It then becomes the reponsability of the Linux users to protect their files, using, for example, encrypted disks.
-
+l
 #### Current API
 
 ```csharp
@@ -58,20 +58,16 @@ If this method fails it throws an exception with more details. Typically the fai
 - D-BUS is not running (typical in SSH scenario)
 - No wallet is listening on the other end
 
-TODO: understand perf impact of this call. If expensive, need to inform consumers to not call this too often.
-
 #### Add a method to persist data in a plaintext file
 
 
 ```csharp
  new StorageCreationPropertiesBuilder(Config.CacheFileName, Config.CacheDir, Config.ClientId) 
-    .WithLinuxUnprotectedFile(Config.UnprotectedCacheFilePath) //new method                     
+    .WithLinuxUnprotectedFile() //new method                     
     .WithMacKeyChain(...); // no change
                      
 ```                     
-If `UnprotectedCacheFilePath` and  `CacheFileName` are the same, an exception MUST be thrown.
-
-As an alternative, the `.WithLinuxPlaintextFile()` can take 0 arguments, and simply use `Config.CacheFileName`, however this would invalidate Goal 4.
+`Config.CacheFileName` will contain the unprotected cache. 
 
 #### Suggested pattern for extension consumers
 
@@ -83,3 +79,14 @@ Libraries consuming the extension will:
 4. Create a cache helper using `.WithLinuxPlaintextFile` using a file path that comes from either: 
 - a well known env variable, e.g. LINUX_DEV_TOOLS_TOKEN_CACHE
 - if LINUX_DEV_TOOLS_TOKEN_CACHE is not set, default to a well known location 
+
+
+#### Important note about signaling API
+
+Some consumenrs of the library are using the event `CacheChanged`. While not all extensions expose this event, all extension need to ensure the event is triggered. 
+This is done via a `FileWatcher` mechanism as follows:
+
+- on Windows, encrypted data is persisted to `Config.CacheFileName` 
+- on Mac, data is stored in KeyChain. A dummy 1 byte is written to `Config.CacheFileName`
+- on Linux, data is persisted EITHER in KeyRin or in `Config.CacheFileName`. To maintain the signaling semantics, a dummy 1 byte will be written to a file named
+`Config.CacheFileName` + '.signal'
--- a/docs/keyring_fallback_proposal.md
+++ b/docs/keyring_fallback_proposal.md
@@ -1,0 +1,85 @@
+# Proposal for Linux plain-text file fallback
+
+**Context:** Today, the MSAL extension libraries (.net, java, python, javascript) store secrets in KeyRings via LibSecret. These components not available on all Linux distros and they cannot be started when connected via SSH.
+
+**Proposal:** Extensions are to provide a mechanism for products to detect if this secret storage is usable. If it is not, extensions are to write the token cache to an unecrypted file. It then becomes the reponsability of the Linux users to protect their files, using, for example, encrypted disks.
+
+#### Current API
+
+```csharp
+ var storagePropertiesBuilder =
+    // CacheFileName is used for storgae only on Win. On Mac and Linux, it's used 
+    // to produce an event, but contents are empty. Actual secrets are stored in KeyRing 
+    // KeyChain.
+    new StorageCreationPropertiesBuilder(Config.CacheFileName, Config.CacheDir, Config.ClientId)
+        .WithLinuxKeyring(
+            Config.LinuxKeyRingSchema,
+            Config.LinuxKeyRingCollection,
+            Config.LinuxKeyRingLabel,
+            Config.LinuxKeyRingAttr1,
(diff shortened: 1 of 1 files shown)


docs/keyring_fallback_proposal.md
```diff
@@ -0,0 +1,85 @@
...
+            Config.LinuxKeyRingSchema,
+            Config.LinuxKeyRingCollection,
+            Config.LinuxKeyRingLabel,
+            Config.LinuxKeyRingAttr1,
+            Config.LinuxKeyRingAttr2)
+        .WithMacKeyChain(
+            Config.KeyChainServiceName,
+            Config.KeyChainAccountName);
+
+MsalCacheHelper cacheHelper = MsalCacheHelper.Create(storageCreationProperties);
```
henrik-me: >storageCreationProperties [](start = 53, length = 25)

should this be the variable defined above?

docs/keyring_fallback_proposal.md
```diff
@@ -0,0 +1,85 @@
...
+- D-BUS is not running (typical in SSH scenario)
+- No wallet is listening on the other end
+
+TODO: understand perf impact of this call. If expensive, need to inform consumers to not call this too often.
+
+#### Add a method to persist data in a plaintext file
+
+
+```csharp
+ new StorageCreationPropertiesBuier(Config.CacheFileName, Config.CacheDir, Config.ClientId)
```
henrik-me: >StorageCreationPropertiesBuier [](start = 5, length = 30)

spelling

docs/keyring_fallback_proposal.md
```diff
@@ -0,0 +1,85 @@
...
+
+
+```csharp
+ new StorageCreationPropertiesBuier(Config.CacheFileName, Config.CacheDir, Config.ClientId) 
+    .WithLinuxPlaintextFile(ConPlaintextFilePath) //new method                     
+    .WithMacKeyChain(...); // no change
+                     
+```                     
+
+As an alternative, the `.WithLinuxPlaintextFile()` can take 0 arguments, and simply use `Config.CacheFileName`, however this would invalidate Goal 4.
```
henrik-me: >however this would invalidate Goal 4. [](start = 112, length = 37)

why? I would not think we delete that file. also devs could specify the same path, what happens then?

docs/keyring_fallback_proposal.md
```diff
@@ -0,0 +1,85 @@
...
+#### Suggested pattern for extension consumers
+
+Libraries consuming the extension will: 
+
+1. create a cache helper with a the normal `KeyRing` setup
+2. call `cacheHelper.VerifyLinuxPersistence()`
+2.1. If this throws an exception, show the user a meaningful message / URL to help page to inform them to secure their secrets storage
+2.2. Create a cache helper using `.WithLinuxPlaintextFile` using a file path that comes from either: 
+a. a well known ENV variable, e.g. DEV_TOOLS_TOKEN_CACHE
+b. if DEV_TOOLS_TOKEN_CACHE is not set, default to a well known location
```
henrik-me: am I understanding correctly that it's the consumer of the helper that decides? If yes, I'm fine with this otherwise I would suggest to rethink

docs/keyring_fallback_proposal.md
```diff
@@ -0,0 +1,85 @@
...
+- No wallet is listening on the other end
+
+TODO: understand perf impact of this call. If expensive, need to inform consumers to not call this too often.
+
+#### Add a method to persist data in a plaintext file
+
+
+```csharp
+ new StorageCreationPropertiesBuier(Config.CacheFileName, Config.CacheDir, Config.ClientId) 
+    .WithLinuxPlaintextFile(ConPlaintextFilePath) //new method
```
henrik-me: >WithLinuxPlaintextFile [](start = 5, length = 22)

should this highlight "not encrypted" or similar making the API even more clear on the missing encryption at rest?

docs/keyring_fallback_proposal.md
```diff
@@ -0,0 +1,85 @@
+# Proposal for Linux plain-text file fallback
```
henrik-me: CLI and PowerShell should both add telemetry allowing us to see how many users fallback to this

docs/keyring_fallback_proposal.md
bgavrilMS: So that file is used for storage in Windows. On Mac and Linux, we still create the file, [but write a single 0 byte into it](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/master/src/Microsoft.Identity.Client.Extensions.Msal/Os/Linux/CacheAccessorLinux.cs#L138)

So we use that file as a signaling mechanism that "something has changed". On Windows, we use it also to cache the tokens. But I don't see how we can use it to support both scenarios at the same time.

docs/keyring_fallback_proposal.md
bgavrilMS: Yes, I'm proposing that the consumers of the helper (Az CLI, Powershell) to implement the logic to determine the file path.

docs/keyring_fallback_proposal.md
schaabs: Perhaps "Unprotected"?

docs/keyring_fallback_proposal.md
henrik-me: if we can't we should also ensure we throw an exception if these 2 values are the same

bgavrilMS: Yes, good catch.

docs/keyring_fallback_proposal.md
```diff
@@ -0,0 +1,85 @@
...
+
+**Context:** Today, the MSAL extension libraries (.net, java, python, javascript) store secrets in KeyRings via LibSecret. These components not available on all Linux distros and they cannot be started when connected via SSH.
```
rayluo: Actually, the MSAL Python extension did not implement KeyRings/LibSecret support, so, [MSAL Python extension's current behavior on Linux is always "fall back" to a plain text file](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/0.1.3/msal_extensions/__init__.py#L6-L11). The extension would print a warning message so that the end user would be informed about [what's going on](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/0.1.3/msal_extensions/token_cache.py#L110-L112) (in such case, we should have also suggest end user to consider disk encryption, though).

docs/keyring_fallback_proposal.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+This method MUST not affect the token cache. It will attempt to write and read a dummy secret. Different KeyRing attributes will be used so as to not interfere with the real token cache. 
+
+If this method fails it throws an exception with more details. Typically the failure points are:
+
+- LibSecret is not installed
+- Incorrect version of LibSecret is installed
+- D-BUS is not running (typical in SSH scenario)
+- No wallet is listening on the other end
```
rayluo: My newbie question here. Assuming we do not have such a `VerifyLinuxEncryption()` API yet, when will those errors/exceptions be detected/thrown? During `class LinuxKeyRingStorage(...)` construction, or during `storage.Save("...")`? Or would there be no exception at all (gasp!), so that we would need to resort to "write and read a dummy secret (and compare them)"?

docs/keyring_fallback_proposal.md
```diff
@@ -0,0 +1,85 @@
...
+
+As an alternative, the `.WithLinuxPlaintextFile()` can take 0 arguments, and simply use `Config.CacheFileName`, however this would invalidate Goal 4.
+
+#### Suggested pattern for extension consumers
+
+Libraries consuming the extension will: 
+
+1. create a cache helper with a the normal `KeyRing` setup
+2. call `cacheHelper.VerifyLinuxPersistence()`
+2.1. If this throws an exception, show the user a meaningful message / URL to help page to inform them to secure their secrets storage
```
rayluo: Editorial suggestion: Markdown does not support multi-layer order, so it won't [render this paragraph](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/bogavril/keyring_proposal/docs/keyring_fallback_proposal.md#suggested-pattern-for-extension-consumers) properly. Perhaps we can rephrase these, somehow.

docs/keyring_fallback_proposal.md
```diff
@@ -0,0 +1,85 @@
...
+```                     
+
+As an alternative, the `.WithLinuxPlaintextFile()` can take 0 arguments, and simply use `Config.CacheFileName`, however this would invalidate Goal 4.
+
+#### Suggested pattern for extension consumers
+
+Libraries consuming the extension will: 
+
+1. create a cache helper with a the normal `KeyRing` setup
+2. call `cacheHelper.VerifyLinuxPersistence()`
```
rayluo: Could this detection happen inside the constructor of the KeyRing class, or inside a `HelperFactory()` method?

docs/keyring_fallback_proposal.md
bgavrilMS: Is this used in production by anybody? Because our 2 extensions would not work with each other right now.

rayluo: [Some open source packages out there use MSAL Python extension](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/network/dependents?package_id=UGFja2FnZS00NjMyNzUyOTY%3D). We do not have numbers for private projects. Regardless, 3rd-party apps would typically not need/want token sharing with other apps.

IIRC, last time our internal partners were prototyping on Windows. This time we will enable this token sharing scenario on Linux.

bgavrilMS: That's pretty cool. Yeah, only 1st party apps would want to share the token cache. 

I think the interface against which our extension libraries work revolves under the assumption that: 

- data is stored in the same place (KeyChain, KeyRing, DPAPI file)
- the Config.CacheFileName file is written to (potentially with dummy data  if data is stored somewhere else). This signals to other apps that "the cache changed".

docs/keyring_fallback_proposal.md
markcowl: Also, what is the behavior if there is a file system error when creating the cache fallback? Do we create an in-memory cache, or fail?

Assume that the `Validate*` method would throw an exception if
- The backup file cache were used
- an in-memory cache were used

Is that correct?

docs/keyring_fallback_proposal.md
```diff
@@ -0,0 +1,86 @@
...
+
+MsalCacheHelper cacheHelper = MsalCacheHelper.Create(storageProperties);
+cacheHelper.RegisterCache(app.UserTokenCache);
+
+```
+## Goals
+
+1. API for fallback to file for Linux. (P0)
+
+2. Developers must opt-in to fallback, it should not be a default, since fallback is insecure. (P0)
```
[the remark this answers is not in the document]
bgavrilMS: SSO should work when using the file callback, since this is likely to be the default mode for PowerShell and Azure CLI over SSH. Of course, the same file path needs to be used.

docs/keyring_fallback_proposal.md
bgavrilMS: @markcowl - yes, this is my understanding. @henrik-me , @jmprieur ?

docs/keyring_fallback_proposal.md
bgavrilMS: Currently the extension does not throw exceptions. I believe the original intent was to not crash for small one-off errors since apps can continue to work anyway with minor data loss. I'm hesitant to change this behavior since the extension are used in production.

And yes, `ValidatePersistence` would write, read and clear.

docs/keyring_fallback_proposal.md
jmprieur: Yes, absolutely @markcowl 
cc; @henrik-me @bgavrilMS @navyasric @mvrak

rayluo: Yes, "the intention long term is that we will implement the same mechanism on linux in all 4 major languages". My first message in this thread was just to call out that LibSecret support was not prevoiusly implemented in msal python extension. And I heard from @SomkaPe that it is also unimplemented for msal java extension today. So that our planning would need to take these into consideration.

docs/keyring_fallback_proposal.md
rayluo: MSAL Python extension did not absorb those exceptions (on Windows/macOS), so the downstream apps would already have a chance to know the errors. Besides, as I mentioned above, that the libsecret support (on Linux) is not currently available in msal python/java extension. We will define and implement that set of API, and have a better idea on how/whether a `ValidatePersistence()` would fit into the programming model.

docs/keyring_fallback_proposal.md
rayluo: Hi @markcowl , does [this code snippet](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/pull/60#discussion_r377923730) imply an app would use multiple storage AT THE SAME TIME, wrapped inside one super helper? Is that a goal?

We might not have a clear definition on this previously. Currently, msal python extensions provide 3 different flavors, and then the app can choose to use one of them based on different circumstance. Also, the one they choose would simply REPLACE the default in-memory token cache. @bgavrilMS how does that work in C#? @SomkaPe How about java?

markcowl: No, this is just specifying the fallback mechanism - Just as you use MacKeyhain if you are on MacOS, you use the fallback if this code is run on a linux distro without libsecret support.

Given the above code, the client would

- Use MacKeyChain if on MacOS
- Use libesecret if on linux and it is available
- use file fallback if on linux and libsecret is not available

rayluo: Thanks for the clarification. We are now on the same page about the high level goal "multiple persistence options will be provided by msal extensions, but only one persistence would be chosen at run-time".

Comparing to that, the interface of "how-to-choose" could be an implementation-level decision. The builder pattern, in itself, might imply an "AND" relationship among its `WithXyz(...).WithAbc(...)`, without particular order. We can leave this to task owners to explore different options and finalize the proposal during the implementation phase.

markcowl: Correct, we just want to ensure that we are using the same logic to decide on the persistence mechanism for clients opting in to the shared cache on each platform

docs/keyring_fallback_proposal.md
bgavrilMS: @henrik-me  - resolved
@markcowl - we don't want to add the "dev tools" details as defaults. This is an open source implementation of the token cache, and 3rd party developers are slowly picking it up as well. We want to avoid them accidentally using the same locations as the dev tools.

docs/keyring_fallback_proposal.md
bgavrilMS: The application consuming the extension needs to know if the underlying persistence is working or not. I suppose you can ask the application to first use the LibSecret version if that fails during `RegisterCache` to use the fallback.

docs/keyring_fallback_proposal.md
bgavrilMS: I really want to make sure that we show the end-user a very clear message that their token cache will be stored unencrypted. To me this pattern is more clear: 

1. Create `StorageProperties` with `LinuxKeyRing` only
2. Call `ValidatePersistence`
3. If the call fails, tell the user that their cache will be stored unencrypted
4. Create `StorageProperties` with `UnencryptedFile` instead

docs/keyring_fallback_proposal.md
bgavrilMS: Yes, the 2 options are mutually exlcusive.

docs/keyring_fallback_proposal.md
SomkaPe: I would go with second option - by using meaningful method name for "WithLinuxUnprotectedFallback"
we totally can deliver the message.
I personally like very precise naming - so i would use - WithLinuxUnencryptedFileFallback 
- to be consistent with other methods , looks like it follows next format : With[Platform][PersistenceMechanism]

Or may be even WithLinuxUnencryptedFileIfKeyRingNotAvailable

As app developer i would better see this method than wondering what does it mean "unprotected", 
or "fallback"

docs/keyring_fallback_proposal.md
```diff
@@ -0,0 +1,87 @@
...
+#### Suggested pattern for extension consumers
+
+Libraries consuming the extension will: 
+
+1. create a cache helper with a the normal `KeyRing` setup
+2. call `cacheHelper.VerifyLinuxPersistence()`
+3. If this throws an exception, show the user a meaningful message / URL to help page to inform them to secure their secrets storage
+4. Create a cache helper using `.WithLinuxPlaintextFile` using a file path that comes from either: 
+- a well known env variable, e.g. LINUX_DEV_TOOLS_TOKEN_CACHE
+- if LINUX_DEV_TOOLS_TOKEN_CACHE is not set, default to a well known location
```
SomkaPe: i would prefer to get it directly from app developer, so they will be aware about existence of that file.
This would better align with goal 2 "Developers must opt-in to fallback"

docs/keyring_fallback_proposal.md
```diff
@@ -0,0 +1,87 @@
...
+```
+## Goals
+
+1. API for fallback to file for Linux. (P0)
+
+2. Developers must opt-in to fallback, it should not be a default, since fallback is insecure. (P0)
+
+3. API for detecting if persistence is not working. This will allow products to show users a warning message about the fallback. (P1)
+
+4. If a user connects both via SSH and via UI, her SSH token cache (i.e. the plaintext file) should not be deleted. (P2)
```
SomkaPe: deleted by what ? as i understand Msal and MsalExtension never delete cache file

docs/keyring_fallback_proposal.md
SomkaPe: > Also, what is the behavior if there is a file system error when creating the cache fallback? Do we create an in-memory cache, or fail?
> 
> Assume that the `Validate*` method would throw an exception if
> 
> * The backup file cache were used
> * an in-memory cache were used
> 
> Is that correct?

good question !
Should we support in-memory mode ? 
when app encounter I/O failure What should it do ? May be switching to in-memory mode - it  will not be sharing cache with others But it will be possible to use application.
But sing out by other app will be ignored - may be not a big problem since it is just in-memory state

bgavrilMS: @SomkaPe - yes, if I/O failure occurs, in-memory version is used. This is the reason why .net extension does not throw exceptions on I/O failures.

docs/keyring_fallback_proposal.md
bgavrilMS: MSAL Extension in .NET writes a dummy 1 byte to the CacheFile on Mac and Linux. This file is "watched" by a FileWatcher construct, which fires a callback everytime the file changes. MSAL Extension then calculates if accounts were added or removed and produces a callback named

`CacheChanged(IEnumerable<IAccount> accountsAdded, IEnumerable<IAccount> accountsRemoved)`

Now imagine the following scenario: 

1. User connects via SSH, fallback cache is used, we write tokens to CacheFile 
2. User connects via UI, KeyRing is used. Dummy 1 byte is written to cache file. 

Perhaps this is too complex and we should just not support SSH + UI connection. This will simplify things..

docs/keyring_fallback_proposal.md
rayluo: 1. With regard to the In-Memory mode, i.e. "if I/O failure occurs, do not throw exception, just silently switch to in-memory cache", if that is historically already the existing behavior in some of our implementations, there might not be a motivation to change. (By the way, will those msal extensions silently switch back at the next token cache I/O opportunity? If so, what is the merging logic?)

   FWIW, msal python extensions did not have/maintain an in-memory mode, so, there is conceptually only one copy of cache, and that is on-disk. File I/O exception would be bubbled up, but apps can choose to explicitly ignore them. Next cache interaction would read latest content from on-disk cache again.

2. We briefly discussed the API for creating an extension. One alternative might be a factory pattern with explicit error/exception, so that app developers do the explicit opt-in. Something like this:

```python
try:
    cache = CacheFactory(...)
except:
    log.warn("Could not create an encrypted cache, fallback to an unprotected cache")
    cache = PlaintextCacheFile(...)

pca = PublicClientApplication(..., cache=cache)
pca.acquire_token_by_xyz(...)
```

PR #62, opened 2020-02-12
Add a VerifyPersistence API to the cache helper

bgavrilMS:
Add a `VerifyPersistence` convenience API that works by: 

- creates a cache accessor similar to the "main" one (not identical, otherwise it will overwrite the cache)
- writes, reads and clears some data
- throws an exception if the above fails

Tested on Windows (happy path only) and Linux - no libsecret and libsecret installed but not working (SSH). 

Testing remaining: Mac (happy path), Linux (happy path)

### DIFF
--- a/src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheStorage.cs
+++ b/src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheStorage.cs
@@ -187,19 +187,18 @@ namespace Microsoft.Identity.Client.Extensions.Msal
             }
         }
 
-        /// <summary>
-        /// 
-        /// </summary>
         public void VerifyPersistence()
         {
-
             try
             {
+                // do not use the _cacheAccessor for writing dummy data, as it might overwrite an actual token cache
+                var persitenceValidatationAccessor = _cacheAccessor.CreateForPersistenceValidation();
+
                 _logger.LogInformation($"[Verify Persistence] Writing Data ");
-                _cacheAccessor.Write(Encoding.UTF8.GetBytes(PersistenceValidationDummyData));
+                persitenceValidatationAccessor.Write(Encoding.UTF8.GetBytes(PersistenceValidationDummyData));
 
                 _logger.LogInformation($"[Verify Persistence] Reading Data ");
-                var data = _cacheAccessor.Read();
+                var data = persitenceValidatationAccessor.Read();
 
                 if (data == null || data.Length == 0)
                 {
@@ -217,7 +216,7 @@ namespace Microsoft.Identity.Client.Extensions.Msal
                 }
 
                 _logger.LogInformation($"[Verify Persistence] Clearing data");
-                _cacheAccessor.Clear();
+                persitenceValidatationAccessor.Clear();
             }
             catch (Exception ex) when (!(ex is MsalCachePersistenceException))
             {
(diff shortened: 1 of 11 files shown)


tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/AssertException.cs
```diff
@@ -0,0 +1,243 @@
...
+using System;
+using System.Diagnostics;
+using System.Globalization;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Identity.Client.Extensions.Msal.UnitTests
+{
+    public static class AssertException
```
bgavrilMS: Copied over from MSAL, pls don't review.

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheStorage.cs
```diff
@@ -185,5 +186,43 @@ public void Clear()
...
+                persitenceValidatationAccessor.Write(Encoding.UTF8.GetBytes(PersistenceValidationDummyData));
+
+                _logger.LogInformation($"[Verify Persistence] Reading Data ");
+                var data = persitenceValidatationAccessor.Read();
+
+                if (data == null || data.Length == 0)
+                {
+                    throw new MsalCachePersistenceException(
+                        "Persistence check failed. Data written could not be read. " +
+                        "Possible cause: on Linux, LibSecret is installed by D-Bus isn't running because it cannot be started over SSH.");
```
erich-wang: typo ? LibSecret is installed **by** D-Bus

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheStorage.cs
```diff
@@ -185,5 +186,43 @@ public void Clear()
...
+                string dataRead = Encoding.UTF8.GetString(data);
+                if (!string.Equals(PersistenceValidationDummyData, dataRead, StringComparison.Ordinal))
+                {
+                    throw new MsalCachePersistenceException(
+                        $"Persistence check failed. Data written {PersistenceValidationDummyData} is different from data read {dataRead}");
+
+                }
+
+                _logger.LogInformation($"[Verify Persistence] Clearing data");
+                persitenceValidatationAccessor.Clear();
```
erich-wang: suggest to put line 219 in finally to make sure it is called even line 213 happens

bgavrilMS: Agreed, thx.

bgavrilMS:
Thanks @erich-wang , will address your comments and commit.

src/Microsoft.Identity.Client.Extensions.Msal/Os/Linux/CacheAccessorLinux.cs
```diff
@@ -9,33 +9,94 @@ namespace Microsoft.Identity.Client.Extensions.Msal
...
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
         }
 
+        public ICacheAccessor CreateForPersistenceValidation()
+        {
+            return new CacheAccessorLinux(
+                _cacheFilePath,
+                _keyringCollection,
+                _keyringSchemaName,
+                "MSAL Test",
```
henrik-me: >"MSAL Test" [](start = 16, length = 11)

MSAL Persistence Test?

henrik-me:
Does these tests run on a linux box and a mac box?

bgavrilMS:
These tests indeed run on a mac box, not a linux box afaik, but since the actual accessors are mocked, it doesn't matter that much.

PR #66, opened 2020-02-21
Remove MsalCacheStorage from public API and expose Import / Export at higher level

bgavrilMS:


bgavrilMS:
@erich-wang - please have a look at this PR

Issue #61, opened 2020-02-11
PlatformNotSupportedException on MacOS via Xamarin / Make CacheChanged event configurable


tipa:
I tried to implement token serialization as described [here](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/wiki/Cross-platform-Token-Cache
) for my Xamarin.MacOS app. However, I am hitting this exception:

```
System.PlatformNotSupportedException: Operation is not supported on this platform.
  at System.IO.FileSystemWatcher.StartRaisingEvents () [0x00000] in /Library/Frameworks/Xamarin.Mac.framework/Versions/Current/src/Xamarin.Mac/external/corefx/src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.UnknownUnix.cs:23 
  at System.IO.FileSystemWatcher.StartRaisingEventsIfNotDisposed () [0x00019] in /Library/Frameworks/Xamarin.Mac.framework/Versions/Current/src/Xamarin.Mac/external/corefx/src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.cs:664 
  at System.IO.FileSystemWatcher.set_EnableRaisingEvents (System.Boolean value) [0x0001d] in /Library/Frameworks/Xamarin.Mac.framework/Versions/Current/src/Xamarin.Mac/external/corefx/src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.cs:163 
  at (wrapper remoting-invoke-with-check) System.IO.FileSystemWatcher.set_EnableRaisingEvents(bool)
  at Microsoft.Identity.Client.Extensions.Msal.MsalCacheHelper.CreateAsync (Microsoft.Identity.Client.Extensions.Msal.StorageCreationProperties storageCreationProperties, System.Diagnostics.TraceSource logger) [0x000bc] in <e1855c7c1c93442fa2635ff71c06cc54>:0 
```

My code:
```
var storageProperties = new StorageCreationPropertiesBuilder("msal_cache.dat", Environment.GetFolderPath(Environment.SpecialFolder.Personal), ClientId)
    .WithMacKeyChain("msal_service", "msal_account")
    .Build();
var cacheHelper = await MsalCacheHelper.CreateAsync(storageProperties);
```

bgavrilMS:
[...]

PR #69, opened 2020-03-06
Perf improvements 

bgavrilMS:


src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
```diff
@@ -142,10 +136,16 @@ private MsalCacheHelper(StorageCreationProperties storageCreationProperties, Tra
 
         private async void OnCacheFileChangedAsync(object sender, FileSystemEventArgs args)
         {
+            // avoid the high cost of computing the added / removed accounts if nobody listens to this
+            if (CacheChanged == null)
```
bgavrilMS: Main perf improvement - do not deserialize the cache to figure out which accounts were changed if event is not used.

tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/Resources/token_cache_one_acc_seed.json
```diff
@@ -0,0 +1,57 @@
+﻿{
+    "AccessToken": {
+        "ae821e4d-f408-451a-af82-882691148603.49f548d0-12b7-4169-a390-bb5304d24462-login.windows.net-accesstoken-1d18b3b0-251b-4714-a02a-9956cec86c2d-49f548d0-12b7-4169-a390-bb5304d24462-files.readwrite files.readwrite.all openid profile user.read user.readbasic.all user.readwrite email": {
+            "home_account_id": "ae821e4d-f408-451a-af82-882691148603.49f548d0-12b7-4169-a390-bb5304d24462",
+            "environment": "login.windows.net",
+            "client_info": "eyJ1aWQiOiJhZTgyMWU0ZC1mNDA4LTQ1MWEtYWY4Mi04ODI2OTExNDg2MDMiLCJ1dGlkIjoiNDlmNTQ4ZDAtMTJiNy00MTY5LWEzOTAtYmI1MzA0ZDI0NDYyIn0",
+            "client_id": "1d18b3b0-251b-4714-a02a-9956cec86c2d",
+            "secret": "secret",
```
bgavrilMS: no secrets here :)

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
bgavrilMS: @henrik-me - this event is fired a lot but most consumers do not require it. This quick fix will improve perf, but proposal is to make it configurable, smth like: 

`.WithCacheChangedEvent(true)`

And the default will be `false`. Breaking change, but only VS uses it now, so we should be safe. I'll bump the major version. Does this sound ok?

henrik-me: how about a construct where we don't have to bump the major version?

---
In reply to: [389782812](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/pull/69#discussion_r389782812) [](ancestors = 389782812)

bgavrilMS: We can leave the default to "true". This is fine, but not very discover-able, and users will just leave this turned on, with the added perf implications. 

Alterntively, we can ask for this flag to be a mandatory parameter, but it's still a breaking change :)

henrik-me: yeah... I'm ok with this change as long as VS is ok to opt in and take this update. Also other tools will likely have to opt in. Assuming for those where this makes most impact they will all be opting in (eventually).

Also I believe WithCacheChangedEvent default should be true. at least to me that makes more sense

---
In reply to: [389797301](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/pull/69#discussion_r389797301) [](ancestors = 389797301)

henrik-me: Why opt in? If the handler is not set we can ignore throwing these events. Not seeing how this would be a breaking change - imo this is fixing a performance bug. 

---
In reply to: [389801029](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/pull/69#discussion_r389801029) [](ancestors = 389801029,389797301)

bgavrilMS: Agreed. If the handler is not set, then perf improves quite a bit. However: 

1. We are still spinning off new threads, even though they do almost no work
2. People have complained that the FileWatcher does not work in Xamarin.Mac (separate issue - #61)

I think we're good to go for now, and we can revisit this as part of #61

PR #71, opened 2020-03-25
Run unit tests on both .net classic and .net core

bgavrilMS:


tests/Microsoft.Identity.Client.Extensions.Adal.UnitTests/Microsoft.Identity.Client.Extensions.Adal.UnitTests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFrameworks>netcoreapp3.1;net472</TargetFrameworks>
```
bgavrilMS: It looks like Visual Studio finally understands how to run tests on 2 platforms. We had to do some wierd stuff in MSAL to be able to do this.

tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/ResourceHelper.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// <summary>
+        /// Gets the relative path to a test resource. Resource should be using DeploymentItem (desktop) or
+        /// by setting Copy to Output Directory to Always (other platforms)
+        /// </summary>
+        /// <remarks>
+        /// This is just a simple workaround for DeploymentItem not being implemented in mstest on netcore
+        /// Tests seems to run from the bin directory and not from a TestRun dir on netcore
+        /// Assumes resources are in a Resources dir.
+        /// </remarks>
+        public static string GetTestResourceRelativePath(string resourceName)
```
bgavrilMS: workaround for mstest issue ... we use it in MSAL as well...

PR #72, opened 2020-03-26
Bogavril/fallback

bgavrilMS:
In parallel with this feature I am reviewing all tests so that we have CI running on Win, Mac and Linux (currently it only runs on Win). 

This PR contains the first Linux test and there are a bunch of Mac tests already.


src/Shared/SharedUtilities.cs
```diff
@@ -18,7 +18,7 @@ namespace Microsoft.Identity.Client.Extensions.Web
     /// <summary>
     /// A set of utilities shared between service and client
     /// </summary>
-    internal static class SharedUtilities
+    public static class SharedUtilities
```
bgavrilMS: I decided to expose this utilities class because there are some helpers that are useful to consumers.

tests/ManualTestApp/ExampleUsage.cs
```diff
@@ -13,37 +14,93 @@ namespace ManualTestApp
     /// </summary>
     public class ExampleUsage
```
bgavrilMS: This class shows how to use the library.

tests/ManualTestApp/ExampleUsage.cs
```diff
@@ -13,37 +14,93 @@ namespace ManualTestApp
...
+
+                return cacheHelper;
+            }
+            catch (MsalCachePersistenceException ex)
+            {
+                Console.WriteLine("Cannot persist data securely. ");
+                Console.WriteLine("Details: " + ex);
+
+
+                if (SharedUtilities.IsLinuxPlatform())
```
bgavrilMS: This is where consumers would need the helpers...

tests/ManualTestApp/ExampleUsage.cs
```diff
@@ -13,37 +14,93 @@ namespace ManualTestApp
...
+                {
+                    storageProperties = ConfigureSecureStorage(usePlaintextFileOnLinux: true);
```
bgavrilMS: All in all, the pattern of usage is pretty verbose, but this was a design requirement.

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/UnencryptedFileAccessor.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.Extensions.Msal
+{
+    internal class UnencryptedFileAccessor : ICacheAccessor
```
henrik-me: >UnencryptedFileAccessor [](start = 19, length = 23)

Perhaps simply FileAccessor?

src/Shared/StorageCreationPropertiesBuilder.cs
```diff
@@ -126,12 +128,40 @@ public StorageCreationPropertiesBuilder CustomizeLockRetry(int lockRetryDelay, i
                 throw new ArgumentNullException(nameof(schemaName));
             }
 
+            if (_useLinuxPlaintextFallback)
+            {
+                throw new ArgumentException(
+                    "WithLinuxUnprotectedFile and WithLinuxKeyring are mutually exclusive. " +
+                    "Only one storage mechanism per OS is supported. ");
```
henrik-me: ak.ms link to doc? (same below)?

src/Shared/StorageCreationPropertiesBuilder.cs
```diff
@@ -126,12 +128,40 @@ public StorageCreationPropertiesBuilder CustomizeLockRetry(int lockRetryDelay, i
...
+
+        /// <summary>
+        /// Use to allow storage of secrets in the cacheFile which was configured in the constructor of this class.
+        /// Secrets are stored in PLAINTEXT!
+        /// Should be used as a fallback for cases where Linux LibSecret is not available, for example
+        /// over SSH connections. Users are responsible for security.
+        /// </summary>
+        /// <remarks>You can check if the persistence is available by calling msalCacheHelper.VerifyPersistence() /></remarks>
+        /// <returns></returns>
+        public StorageCreationPropertiesBuilder WithLinuxUnprotectedFile()
```
henrik-me: >WithLinuxUnprotectedFile [](start = 48, length = 24)

I thought the use case was that this at run-time would allow fallback to the unprotected file in case encryption was not available and the developer added WithUnprotectedFile() ? Now the developer will actively have to have two StorageCreationPropertiesBuilder. Can't we guarantee the keyring is used if available or otherwise fallback instead of having the developers do this additional effort? PerhapsWithAllowUnproctedfile ? (or a bool on WithKeyring....). Not sure as we also want to make developers really aware of what they are doing, but if we do the above then we will still try to do most secure for the system....

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/LinuxKeyRingAccessor.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Microsoft.Identity.Client.Extensions.Msal
 {
-    internal class CacheAccessorLinux : ICacheAccessor
+    internal class LinuxKeyRingAccessor : ICacheAccessor
```
pmaytak: nit: 'Keyring', lowercase 'r'. https://wiki.gnome.org/action/show/Projects/GnomeKeyring?action=show&redirect=GnomeKeyring

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/MacKeyChainAccessor.cs
```diff
@@ -9,14 +9,14 @@ namespace Microsoft.Identity.Client.Extensions.Msal
     /// <summary>
     /// 
     /// </summary>
-    internal class CacheAccessorMac : ICacheAccessor
+    internal class MacKeyChainAccessor : ICacheAccessor
```
pmaytak: nit: 'Keychain', lowercase 'c'. https://support.apple.com/en-us/HT204085

PR #75, opened 2020-03-27
Proposal for storing custom API

bgavrilMS:


docs/storing_app_data.md
```diff
@@ -0,0 +1,24 @@
...
+**Context:** Consumers of MSAL / MSAL extension need to securily store additional data for their apps, which also benefit from cross-process syncronization
+
+**Proposal:** MSAL extentsion already handles secure storage and cross process locking, so an API can be exposed
+
+```csharp
+// Reuse the existing CacheHelper object, created based on storage config
+byte[] data = "..."; 
+
+// Read / Writes use file locks!
+cacheHelper.WriteCustomData("custom_data_1", data);
```
rayluo: Having "custom" keyword in API surface feels a little bit strange. What if we do not think all these from MSAL EX's perspective, but rather provide a more generic interface, such as:

```pseudo
class Storage
{
    Storate(namespace)
    {
        ...
    }

    read()
    {...}

    bytes[] write()
    {...}
}
```

which is very close to a generic `File("filename")` interface already.

And, interestingly, MSAL EX itself may or MAY NOT use such generic interface, due to potentially different locking requirement. We can discuss more in meeting.

docs/storing_app_data.md
```diff
@@ -0,0 +1,24 @@
...
+cacheHelper.WriteCustomData("custom_data_1", data);  
+var readData = cacheHelper.ReadCustomData("custom_data_1");
+cacheHelper.Clear("custom_data_1");
+```
+
+The behavior of this depends on the operating system: 
+
+- Windows: use  the same location as the cache, but file name will be `custom_data_1`
+- Mac: use the keychain location with service = `custom_data_1` and same account
+- Linux: write in the keyring with additional attribute `custom_data_1 = 1`
```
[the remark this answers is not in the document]
bgavrilMS: You have a point - I haven't tested this yet. We could use a single attribute: 

key: MSAL_CUSTOM_DATA
value: custom_data_1

We can also try playing with the other libsecret configs if needed, there are so many of them :).

docs/storing_app_data.md
bgavrilMS: Sounds good, the `MsalCacheHelper` object is getting too busy anyway.

The only problem that I see is that developers might mistakely use this API for cache storage. That's why I added the name `additional` to the methods.

docs/storing_app_data.md
```diff
@@ -0,0 +1,24 @@
...
+- Linux fallback: write in plaintext in the same location as the cache, but filename will `custom_data_1`
```
qianwens, line 22: >Linux fallback [](start = 2, length = 14)

If fallback is used, is there any flag return so that client side can give warning to customer? Same for the MSAL.cache.

bgavrilMS: In MSAL.net yes, this is how we designed it. See example usage [here](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/bac27e967509ee0c37037f2d4ebf9cd51d31d168/tests/ManualTestApp/ExampleUsage.cs#L35) - it's a bit verbose but allows you to show a message to the end-user. 

@rayluo @SomkaPe for the other libs

bgavrilMS:
@jmprieur - so what you are saying is to think about future use where we might be required to support smth like "give me all the custom data keys" ?

docs/storing_app_data.md
```diff
@@ -0,0 +1,24 @@
+# Proposal for storing custom secrets
+
+**Context:** Consumers of MSAL / MSAL extension need to securily store additional data for their apps, which also benefit from cross-process syncronization
```
markcowl, line 3:
```suggestion
**Context:** Consumers of MSAL / MSAL extension need to securely store additional data for their apps, which also benefit from cross-process synchronization
```

docs/storing_app_data.md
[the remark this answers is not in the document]
markcowl: We just need to make sure we can have the same uniqueness guarantees on each platform - that 'foo' and 'fooextended' do not match the same data.  Other than that, either mechanism would work for our purposes.

docs/storing_app_data.md
bgavrilMS: @SomkaPe  - what is your opinion on this?

rayluo: Since my comment above, I've also been working on the [Linux Encryption feature for MSAL EX Python](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/44). As a by-product, its underlying *Persistence* layer is now implicitly exposed. And it also has an `is_encrypted` flag, desirable by @qianwens, mentioned [here](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/pull/75/files#r402721647).

Developer would not mistakenly use such lower level Persistence API for high level TokenCache, because the persistence API and the cache API are in different type. They are simply not interchangable.

Such a persistence layer could become a building block of the Service Principle Storage, which in turn would ultimately be designed and owned by apps who maintain the service principles?

Please refer to a concrete usage example to store arbitrary json blob (or any data, for that matter) in an encrypted persistence layer, from the sample 2 snippet currently documented inside [the description of the aforementioned PR](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/pull/44).

Issue #78, opened 2020-04-27
KeyRing and KeyChain should touch the file cache, not overwrite it


bgavrilMS:
1. Setup persistence on Linux with fallback
2. Connect via SSH and login - file cache fallback is used
3. Connect again via Gnome UI. Login - KeyRing is used 
4. Connect again via SSH

Actual: SSO is lost, user needs to re-login
Expected: SSO should not be lost

Solution: instead of overwriting token cache file, just touch it.

bgavrilMS:
Fixed in https://www.nuget.org/packages/Microsoft.Identity.Client.Extensions.Msal/2.10.0-preview

PR #79, opened 2020-04-27
Fix bug #78 - touch cache file on mac and linux

bgavrilMS:


src/Microsoft.Identity.Client.Extensions.Msal/FileIOWithRetries.cs
```diff
@@ -43,11 +55,32 @@ internal static void WriteDataToFile(string filePath, byte[] data, TraceSourceLo
...
+        /// </remarks>
+        internal static void TouchFile(string filePath, TraceSourceLogger logger)
+        {
+            string directoryForCacheFile = GetOrCreateParentDirectory(filePath, logger);
+            logger.LogInformation($"Cache file directory exists. '{Directory.Exists(directoryForCacheFile)}' now touching cache file");
 
             TryProcessFile(() =>
             {
-                File.WriteAllBytes(filePath, data);
+                if (!File.Exists(filePath))
```
rayluo, line 76: :+1: Good catch on this edge case!

And then the `File.SetLastWriteTimeUtc(filePath, DateTime.UtcNow);` could probably be put into an `else` clause, the difference would be really minimal, though.

tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/TraceStringListener.cs
```diff
@@ -3,11 +3,24 @@
...
 using System.Text;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace Microsoft.Identity.Client.Extensions.Msal.UnitTests
 {
     public class TraceStringListener : TraceListener
     {
+        private const string TraceSourceName = "TestSource";
+
+        public static (TraceSource, TraceStringListener) Create()
```
pmaytak, line 14: Value tuples are pretty cool. :)

tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/TraceStringListener.cs
```diff
@@ -3,11 +3,24 @@
...
+        {
+            var logger = new TraceSource(TraceSourceName, SourceLevels.All);
+            var listner = new TraceStringListener();
```
pmaytak, line 17: typo: `listner` > `listener`

PR #82, opened 2020-04-29
Test lock under stress

bgavrilMS:


tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/CrossPlatLockTests.cs
```diff
@@ -0,0 +1,101 @@
...
+        }
+    
+        private void ValidateResult(string protectedFile, int expectedNumberOfOperations)
+        {
+            Trace.WriteLine("Protected File Content:");
+            Trace.WriteLine(File.ReadAllText(protectedFile));
+
+            var lines = File.ReadAllLines(protectedFile);
+
+            string previousThread = null;
```
rayluo, line 74: Just want to double check, were you using multi-threads, or multi-processes, during such test? I thought we were going to use multi-processes to simulate the real world situation.

bgavrilMS: Multi-process. That's a bad variable name and I'll change it.

bgavrilMS:
@rayluo - credit goes to @SomkaPe - that's quite an ingenious way of testing the locking.

PR #83, opened 2020-05-01
Fix for #61 - ignore CacheChange event on Xamarin Mac

bgavrilMS:
fix for #61


tipa:
Awesome, thanks for the fix - looking forward to the release on NuGet :)

bgavrilMS:
Thanks @tipa, I just need to test it first to avoid embarrassment :)

tipa:
Sure.. I am using the library with the line commented out (instead of catching the exception) without any problems for around 2 months already if that helps your confidence releasing the update :)

bgavrilMS:
It does, I'll cut a release now.

Issue #81, opened 2020-04-28
Corrupt cache should be deleted


crmann1:
Currently if decryption of the cache fails we delete the cache.  The same should be true if the cache cannot be deserialized. 

If the cache is not deleted then using the cache will cause the public client application to not function since it can never load it off disk and just throws an exception . 

For example take a cache, and corrupt it, or just put a space in the file. You should see this exception. You will continue to get this exception until the cache is manually deleted. 

 'Microsoft.Identity.Client.MsalClientException: MSAL V3 Deserialization failed to parse the cache contents. Is this possibly an earlier format needed for DeserializeMsalV2? (See https://aka.ms/msal-net-3x-cache-breaking-change)
   at IDictionary<string, JToken> Microsoft.Identity.Client.Cache.TokenCacheJsonSerializer.Deserialize(byte[] bytes, bool clearExistingCacheData)
   at void Microsoft.Identity.Client.TokenCache.Microsoft.Identity.Client.ITokenCacheSerializer.DeserializeMsalV3(byte[] msalV3State, bool shouldClearExistingCache)
   at Task<HashSet<string>> Microsoft.Identity.Client.Extensions.Msal.MsalCacheHelper.GetAccountIdentifiersAsync(StorageCreationProperties storageCreationProperties)+(TokenCacheNotificationArgs args) => { }
   at async Task Microsoft.Identity.Client.TokenCache.Microsoft.Identity.Client.ITokenCacheInternal.OnBeforeAccessAsync(TokenCacheNotificationArgs args)
   at async Task Microsoft.Identity.Client.Cache.CacheSessionManager.RefreshCacheForReadOperationsAsync(TokenTypes cacheEventType)
   at async Task<IEnumerable<IAccount>> Microsoft.Identity.Client.Cache.CacheSessionManager.GetAccountsAsync(string authority)
   at async Task<IEnumerable<IAccount>> Microsoft.Identity.Client.ClientApplicationBase.GetAccountsAsync()
   at async Task<HashSet<string>> Microsoft.Identity.Client.Extensions.Msal.MsalCacheHelper.GetAccountIdentifiersAsync(StorageCreationProperties storageCreationProperties)
   at async Task<MsalCacheHelper> Microsoft.Identity.Client.Extensions.Msal.MsalCacheHelper.CreateAsync(StorageCreationProperties storageCreationProperties, TraceSource logger)

jongio:
[...]

PR #87, opened 2020-05-19
Fix for bug #81 - delete corrupted cache and bump to msal 4.14

bgavrilMS:
#81

### DIFF
--- a/src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
+++ b/src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
@@ -88,7 +88,9 @@ namespace Microsoft.Identity.Client.Extensions.Msal
         /// Gets the current set of accounts in the cache by creating a new public client, and
         /// deserializing the cache into a temporary object.
         /// </summary>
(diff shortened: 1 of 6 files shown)


src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
```diff
@@ -97,10 +99,20 @@ private static async Task<HashSet<string>> GetAccountIdentifiersAsync(StorageCre
...
                 {
-                    var tempCache = MsalCacheStorage.Create(storageCreationProperties, s_staticLogger.Value.Source);
-                    // We're using ReadData here so that decryption is gets handled within the store.
-                    var data = tempCache.ReadData();
-                    args.TokenCache.DeserializeMsalV3(data);
+                    MsalCacheStorage tempCache = null;
+                    try
+                    {
+                        tempCache = MsalCacheStorage.Create(storageCreationProperties, s_staticLogger.Value.Source);
+                        // We're using ReadData here so that decryption is gets handled within the store.
```
pmaytak, line 106:
```suggestion
                        // We're using ReadData here so that decryption is handled within the store.
```

PR #100, opened 2020-09-23
should use test cache file for verifying persistence on Linux

erich-wang:
In LinuxKeyRingAccessor::CreateForPersistenceValidation, we should use msal.cache.test to verify persistence functionality as other platforms, otherwise the real token cache file is deleted after calling VerifyPersistence.

### DIFF
--- a/src/Microsoft.Identity.Client.Extensions.Msal/Accessors/LinuxKeyRingAccessor.cs
+++ b/src/Microsoft.Identity.Client.Extensions.Msal/Accessors/LinuxKeyRingAccessor.cs
@@ -70,7 +70,7 @@ namespace Microsoft.Identity.Client.Extensions.Msal
         public ICacheAccessor CreateForPersistenceValidation()
         {
             return new LinuxKeyringAccessor(
-                _cacheFilePath,
+                _cacheFilePath + ".test",
                 _keyringCollection,
                 _keyringSchemaName,
                 "MSAL Persistence Test",
--- a/tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/MsalCacheStorageIntegrationTests.cs
+++ b/tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/MsalCacheStorageIntegrationTests.cs
@@ -104,10 +104,23 @@ namespace Microsoft.Identity.Client.Extensions.Msal.UnitTests
             AssertException.Throws<MsalCachePersistenceException>(
                 () => store.VerifyPersistence());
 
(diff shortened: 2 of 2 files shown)


henrik-me (approved):
:shipit:

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/LinuxKeyRingAccessor.cs
```diff
@@ -70,7 +70,7 @@ internal class LinuxKeyringAccessor : ICacheAccessor
         public ICacheAccessor CreateForPersistenceValidation()
         {
             return new LinuxKeyringAccessor(
-                _cacheFilePath,
+                _cacheFilePath + ".test",
```
henrik-me, line 73: >_cacheFilePath + ".test", [](start = 16, length = 25)

unit test?

rayluo: I can also attest that this keyring validation is required at run-time, even in production. In that sense, it is less about "unit test", but more like a "trial run" or "dry run".

FWIW, [MSAL Python extensions chose the term "trial_run"](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/0.3.0/msal_extensions/persistence.py#L256-L258) <sub>but then MSAL Python Extensions does not create a file during trial_run, so it is not susceptible to this bug in the first place. That is a refactor idea to pursue, perhaps at a later time.</sub>.

rayluo (approved):
LGTM. Added some minor comment below.

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/LinuxKeyRingAccessor.cs
SomkaPe: it is linux Key Ring accessor, it should not write anything to cache file, @erich-wang could you please provide details about scenario where "cache file is deleted " ?

erich-wang: I updated the integration test to validate the fix.

@SomkaPe , the issue happens on the Linux machine without Key Ring:
- Sdk user first try to use KeyRing by calling VerifyPersistence, the expected exception is thrown
- Fallback to use plain text, token file is created
- Another sdk user try to use KeyRing by calling VerifyPersistence, the token file created in previous step is deleted.

trwalke:
Thanks a lot for this @erich-wang 

Ill merge this in and ship

PR #104, opened 2020-10-15
Add sonar to yaml

keegan-caruso:


azure-pipelines.yml
```diff
@@ -2,35 +2,65 @@ trigger:
...
-  inputs:
-    command: 'build'
-
-- task: DotNetCoreCLI@2
-  displayName: 'Run unit tests'
-  inputs:
-    command: test
-    projects: 'tests/**/*UnitTests*.csproj'
-    arguments: '--no-build --no-restore --collect "Code coverage"'
+jobs:
```
bgavrilMS, line 10: This is great @keegan-caruso ! A few questions: 

1. Do we want SonarQube analysis to happen on PR builds ? Or just CI builds? 
1. Do the Windows build need to run twice?
1. I guess that there is no way to feed full code coverage info in SonarQube, i.e. for the tests that only run on Mac and Linux? I can look at this separately, no worries.

bgavrilMS (approved):
Please see comments first.

azure-pipelines.yml
keegan-caruso: 1) Good question. Do we want the earliest feedback as part of the code review or do we want the metrics to monitor separate from PRs? How you want to leverage the tool?
2) It doesn't need to. This was originally just to keep it logically separated. We could remove windows from the matrix strategy.
3) I'm not entirely sure, I believe you can feed in multiple reports to consider for coverage e.g sonar.cs.vscoveragexml.reportsPaths or sonar.cs.opencover.reportsPaths accept comma delimited or wildcard paths

bgavrilMS: If SonarQube distinguishes between PR and CI builds, then we're good to go. Otherwise, we should probably enable it only on CI builds.

How did you do it with Wilson? Both builds?

keegan-caruso: For Wilson, yes, we do both. More specifically, we defined the sonar build as a separate build that runs on PR and CI

Sonar's model is slightly different, it is based on branches more than build types.

PR builds will show the delta between it and the main branch, CI will be the main metrics / become the new baseline.

Wilson example [Sonar](https://sonarcloud.io/dashboard?id=AzureAD_azure-activedirectory-identitymodel-extensions-for-dotnet&pullRequest=1535) and [PR](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1535)

We aren't seeing a delta in this PR since the default branch isn't seeded with a result yet.

PR #109, opened 2020-10-29
Fix file locking on Mono

mrward:
Mono behaves differently to dotnet and .NET on Windows with file
locks. Creating a FileStream with FileShare.None will not lock the
file when running on Mono. For this to work on Mono FileStream.Lock
has to be called.

Added Mono platform detection and the CrossPlatLock will call
FileStream.Lock on Mono. Also using FileOptions.DeleteOnClose is
a problem on Mono since the file can be deleted from another process
without a lock so the lock file is not deleted when on Mono.

### DIFF
--- a/src/Shared/CrossPlatLock.cs
+++ b/src/Shared/CrossPlatLock.cs
@@ -51,8 +51,24 @@ namespace Microsoft.Identity.Client.Extensions.Web
                         fileShare = FileShare.Read;
                     }
 
-                    fileStream = new FileStream(lockfilePath, FileMode.OpenOrCreate, FileAccess.ReadWrite, fileShare, defaultBufferSize, FileOptions.DeleteOnClose);
-                    
+                    var fileOptions = FileOptions.DeleteOnClose;
+                    if (SharedUtilities.IsMonoPlatform())
+                    {
+                        // Deleting on close/dispose would cause a file locked by another process to be deleted when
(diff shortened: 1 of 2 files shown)


bgavrilMS (approved):
LGTM

tests/FileLockApp/FileLockApp.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFrameworks>netcoreapp3.1;net472</TargetFrameworks>
```
bgavrilMS, line 5: This crashes the build on Mac and Linux, as net472 isn't present there.

```suggestion
   <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
    <TargetFrameworks Condition="$([MSBuild]::IsOsPlatform('Windows'))">$(TargetFrameworks)net472</TargetFrameworks>
        
```

mrward: Might be easier if I just remove this commit if it is causing problems - which it seems to be with the build errors. It is not a big problem to add the other framework to the csproj manually to test on other platforms if needed.

mrward: Removed the file lock app change.

PR #126, opened 2021-02-08
KeyChain API update

bgavrilMS:
Why? 

PowerShell are encountering key chain access bugs, which are hard to reproduce and diagnose. This change might help, as old keychain APIs are deprecated. This also brings more flexibility in case we want to add more features like KeyChain sharing.

Manual test done: 

- login with ManualTestApp using old version of keychain accessor
- upgrade ManualTestApp to new version
- silent login

Pls do not review the native mac code bindings. They were copied from Git Credential Manager, which is maintained by Microsoft / GitHub.


bgavrilMS:
CC @mjcheetham

src/Shared/Mac/MacKeyChain.cs
```diff
@@ -0,0 +1,303 @@
...
+                        {
+                            return Marshal.PtrToStringAuto(buffer, stringLength);
+                        }
+                    }
+
+                    if (CFGetTypeID(value) == CFDataGetTypeID())
+                    {
+                        int length = CFDataGetLength(value);
+                        IntPtr ptr = CFDataGetBytePtr(value);
+                        return Marshal.PtrToStringAuto(ptr, length);
```
bgavrilMS, line 275: There's an Argument exception here because if the "password" is empty, the ptr is IntPtr.Zero and PtrToStringAuto fails.

src/Microsoft.Identity.Client.Extensions.Adal/AdalCacheStorage.cs
```diff
@@ -170,7 +171,14 @@ private byte[] ReadDataCore()
             else if (SharedUtilities.IsMacPlatform())
             {
                 _logger.LogInformation($"ReadDataCore, Before reading from mac keychain");
-                fileData = MacKeyChain.RetrieveKey(CreationProperties.MacKeyChainServiceName, CreationProperties.MacKeyChainAccountName, _logger);
+                MacOSKeychain keychain = new MacOSKeychain();
+                var entry = keychain.Get(
+                    CreationProperties.MacKeyChainServiceName,
+                    CreationProperties.MacKeyChainAccountName);
+
+                if (entry?.Password != null)
```
henrik-me, line 179: previous code assumed no pwd. is this going to breaking?

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/MacKeyChainAccessor.cs
```diff
@@ -45,26 +50,28 @@ public void Clear()
...
-            _logger.LogInformation("ReadDataCore");
-
             _logger.LogInformation($"ReadDataCore, Before reading from mac keychain");
-            byte[] fileData = MacKeyChain.RetrieveKey(_keyChainServiceName, _keyChainAccountName, _logger);
-            _logger.LogInformation($"ReadDataCore, read '{fileData?.Length}' bytes from the keychain");
+            var entry = _keyChain.Get(_service, _account);
+            _logger.LogInformation($"ReadDataCore, After reading mac keychain {entry?.Password?.Length ?? 0} chars");
 
-            return fileData;
+            if (!string.IsNullOrEmpty(entry?.Password))
```
henrik-me, line 64: seems like a breaking change to me?

henrik-me (dismissed):
Have you validated reading the old adal format and msal formats are possible using this? If you have done proper regression testing this might be ok, otherwise I'm seeing minor things like what you also point out about the pwd which could be breaking.

src/Shared/Shared.projitems
```diff
@@ -12,14 +12,21 @@
...
+    <Compile Include="$(MSBuildThisFileDirectory)Mac\MacKeyChain.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)Mac\MacOSKeychainCredential.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)Mac\SecurityFramework.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)SharedUtilities.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)StorageCreationProperties.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)StorageCreationPropertiesBuilder.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)TraceSourceLogger.cs" />
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="$(MSBuildThisFileDirectory)Linux\" />
```
mjcheetham, line 30: You probably don't need this explicit folder entry since you have items already in `Linux`

tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/MacKeyChainTests.cs
```diff
@@ -17,13 +17,15 @@ public class MacKeyChainTests
...
+        private const string TestNamespace = "msal-test";
 
+        MacOSKeychain _macOSKeychain;
         [TestInitialize]
         public void TestInitialize()
         {
             if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
             {
-                MacKeyChain.DeleteKey(ServiceName, AccountName);
+                _macOSKeychain = new MacOSKeychain();
```
mjcheetham, line 28: Not passing in the `TestNamespace` here on purpose?

src/Shared/Mac/MacOSKeychainCredential.cs
```diff
@@ -0,0 +1,28 @@
...
+            Label = label;
+        }
+
+        public string Service { get; }
+
+        public string Account { get; }
+
+        public string Label { get; }
+
+        public string Password { get; }
```
mjcheetham, line 24: To reduce some of the churn, you could make `Password` a `byte[]` and leave it to the caller to turn their data in to bytes, rather than in the `MacOSKeyhain` type.

src/Shared/Mac/MacKeyChain.cs
```diff
@@ -0,0 +1,303 @@
...
+                        int stringLength = (int)CFStringGetLength(value);
+                        int bufferSize = stringLength + 1;
+                        buffer = Marshal.AllocHGlobal(bufferSize);
+                        if (CFStringGetCString(value, buffer, bufferSize, CFStringEncoding.kCFStringEncodingUTF8))
+                        {
+                            return Marshal.PtrToStringAuto(buffer, stringLength);
+                        }
+                    }
+
+                    if (CFGetTypeID(value) == CFDataGetTypeID())
```
mjcheetham, line 271: If you move to making `MacOSKeychainCredential::Password` a `byte[]` rather than a string, you can drop these `CFGetTypeID` checks in this `GetStringAttribute` method, and push the `CFData` handling in to it's own `GetDataAttribute` method that is only used on the `kSecValueData` attribute.

src/Shared/Mac/MacKeyChain.cs
```diff
@@ -0,0 +1,303 @@
...
+            byte[] bytes = Encoding.UTF8.GetBytes(str);
+            return CFStringCreateWithBytes(IntPtr.Zero,
+                bytes, bytes.Length, CFStringEncoding.kCFStringEncodingUTF8, false);
+        }
+
+        private static MacOSKeychainCredential CreateCredentialFromAttributes(IntPtr attributes)
+        {
+            string service = GetStringAttribute(attributes, kSecAttrService);
+            string account = GetStringAttribute(attributes, kSecAttrAccount);
+            string password = GetStringAttribute(attributes, kSecValueData);
```
mjcheetham, line 243: Again, can make this return a `byte[]`:

```suggestion
            byte[] password = GetDataAttribute(attributes, kSecValueData);
```

src/Microsoft.Identity.Client.Extensions.Msal/Microsoft.Identity.Client.Extensions.Msal.csproj
```diff
@@ -7,6 +7,7 @@
     <!--This will generate AssemblyVersion, AssemblyFileVersion and AssemblyInformationVersion-->
     <Version>$(ClientSemVer)</Version>
 
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
mjcheetham, line 10: Do you need this? I don't think you're using raw pointers anywhere?

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/MacKeyChainAccessor.cs
mjcheetham: If you replace the `string MacOSKeychainCredential::Password` property with a `byte[]` one then you can avoid the extra UTF8 string/byte conversions in the accessor.

src/Microsoft.Identity.Client.Extensions.Adal/Microsoft.Identity.Client.Extensions.Adal.csproj
```diff
@@ -20,6 +20,7 @@
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
     <PackageProjectUrl>https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet</PackageProjectUrl>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
mjcheetham, line 23: Again, probably not needed?

src/Shared/Mac/SecurityFramework.cs
```diff
@@ -0,0 +1,219 @@
...
+            kSecAttrLabel = GetGlobal(Handle, "kSecAttrLabel");
+            kSecAttrAccount = GetGlobal(Handle, "kSecAttrAccount");
+            kSecAttrService = GetGlobal(Handle, "kSecAttrService");
+            kSecValueRef = GetGlobal(Handle, "kSecValueRef");
+            kSecValueData = GetGlobal(Handle, "kSecValueData");
+            kSecReturnData = GetGlobal(Handle, "kSecReturnData");
+        }
+
+        [DllImport(SecurityFrameworkLib, CharSet = CharSet.Ansi, CallingConvention = CallingConvention.Cdecl)]
+        public static extern int SessionGetInfo(int session, out int sessionId, out SessionAttributeBits attributes);
```
mjcheetham, line 54: Don't think you need this external method(!) 😉

Might want to go around trimming unused bindings.

mjcheetham (commented):
Looks good, but you can avoid some churn/adapt the code to fit the usage of operating on `byte[]` secrets rather than `string` passwords.

src/Microsoft.Identity.Client.Extensions.Adal/Microsoft.Identity.Client.Extensions.Adal.csproj
bgavrilMS: image

This is the offending code

bgavrilMS: error CS0227: Unsafe code may only appear if compiling with /unsafe

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/MacKeyChainAccessor.cs
bgavrilMS: @henrik-me - it is not a breaking change. The old APIs were more like "ReadPassword" and the new APIs are "ReadEntry", where an entry has Password, Account, Service and Label. We only use the password field.

Btw, I tested this change on old version vs new version.

src/Microsoft.Identity.Client.Extensions.Msal/Microsoft.Identity.Client.Extensions.Msal.csproj
bgavrilMS: needed.

bgavrilMS: Actually, we don't seem to be using `SecKeychainItemCopyAttributesAndData` anywhere, so it should be fine,.

src/Microsoft.Identity.Client.Extensions.Adal/AdalCacheStorage.cs
bgavrilMS: No, previous code would only use the password. This code uses full KeyChain entry structure (account, service, password, label). We only use the password from it.

tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/MacKeyChainTests.cs
bgavrilMS: Yes. Product code doesn't use it, so I'd rather have some test not using it either.

Issue #123, opened 2021-02-01
MsalCacheHelper.CreateAsync throws HttpRequestException when offline


FreddyD-GH:
**Desired behavior**
I would like to be able to read the accounts from the cache, even when the user does not have an internet connection.  It is my understanding that this functionality works when using MSAL with token caches on other platforms.  Is there any reason offline functionality should not be possible with the token cache provided in this repo?

**Actual behavior**
When I call `MsalCacheHelper.CreateAsync`, it throws a `HttpRequestException` with the message `No such host is known`.

**Miscellaneous**
- The app I am using is a .NET 5.0 WPF app.
- I am using version 2.16.7 of Microsoft.Identity.Client.Extensions.Msal

bgavrilMS:
What is your authority? 

This is a valid bug, generally MSAL is able to perform GetAccount offline. As long as you connect to some known clouds, this should work offline. MSAL needs to go online if the are tokens in the cache for authorities it does not recognize.

FreddyD-GH:
I am using `.WithB2CAuthority()` with the form `xxx.b2clogin.com`.

Issue #121, opened 2021-01-26
MsalCacheHelper.Clear() method - change behavior instead of deprecation


rabenshm:
Following offline talk with Bogdan, we agreed that it is better instead of deprecating the MsalCacheHelper.Clear() method, to change its behavior that it will only clear the cache entries relevant to the specific public client application id that the MsalCacheHelper object was created with, and not whole entries in the cache file itself.

bgavrilMS:
Problem is that MsalCacheHelper is constructed with a `client_id` however Clear() method simply removes the data file, i.e. clears the cache for all client_id.

bgavrilMS:
I will have removed the need to pass in the client_id to the cache with the associated PR. So with next release - 2.18.0 - you can have a single instance of `MsalCacheHelper`. 

To sign out users, you need to use MSAL i.e. call `RemoveAccountAsync`. Wiping out the token cache file is discouraged, we will only keep the method for backwards compatibility reasons.

Ping me to discuss further.

Issue #102, opened 2020-10-13
[Feature Request] Cache Helper object should not need client ID  


bgavrilMS:
Currently, the top level object requires to be created by passing in a client ID. Internally this is only used to produce the `CacheChanged` event, which many consumers do not care about. 

We should allow creation of this top level object without a client ID.

Issue #89, opened 2020-07-06
persistent cache resiliency


SomkaPe:
Taking to account that file locks are not completely reliable on non Win platforms, It would be nice to implement following features to be race conditions resilient:
- Switch file cache accessor from non-atomic file writes to atomic file renaming (in order to prevent dirty reads + dirty writes - cache corruption)
- Retry logic on KeyChainAccessor to deal with race conditions

details can be found here -> https://github.com/AzureAD/microsoft-authentication-extensions-for-java/wiki/Race-conditions

henrik-me:
@bgavrilMS : Java and soon Python has these improvements. We should ensure to properly plan for these in .NET.

PR #130, opened 2021-02-11
Fix for #98 - cache cannot be used with ADFS

### LINKED ISSUE #98: MsalCacheHelper.CreateAsync throws MsalClientException when using an ADFS ClientId
When trying to use `MsalCacheHelper` with a ClientId for ADFS which is not a GUID the followoing exception is thrown.

```
MSAL.NetCore.4.16.1.0.MsalClientException:
ErrorCode: client_id_must_be_guid
Microsoft.Identity.Client.MsalClientException: Error: ClientId is not a Guid.
  at Microsoft.Identity.Client.AbstractApplicationBuilder1.Validate() 
  at Microsoft.Identity.Client.PublicClientApplicationBuilder.Validate() 
  at Microsoft.Identity.Client.Extensions.Msal.MsalCacheHelper.GetAccountIdentifiersAsync(StorageCreationProperties storageCreationProperties, TraceSourceLogger logger) 
  at Microsoft.Identity.Client.Extensions.Msal.MsalCacheHelper.CreateAsync(StorageCreationProperties storageCreationProperties, TraceSource logger)
```

To reproduce this you can run the following.

``` c#
StorageCreationProperties storageProperties = new StorageCreationPropertiesBuilder("msal.cache", ".", "AdfsNonGuidClientId").Build();

await MsalCacheHelper.CreateAsync(storageProperties).ConfigureAwait(false);
```

To work around this issue I'm updating the code to supply a random guid rather than a specific client id. The client id doesn't seem to be necessary as the `MsalCacheHelper` only uses it to create a `PublicClientApplication` to iterate through the accounts in a serialized cache, in the method `GetAccountIdentifiersAsync`

https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/1a2de3a3ae272f288dbeda8fc242e7846892848f/src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs#L91-L127

bgavrilMS:
Before this fix, developers had to configure client_id. 

This design does not take into account the following, which this PR fixes:

- it's not clear that `CacheChanged` event is only relevant for the given client_id
- since authority is not configured, the default authority is implied (i.e. `l.m.o/common`). This is the root cause of bug #89 
- developers cannot use 1 instance of the CacheExtension for multiple client_ids (since a client_id is needed to construct it)

Fixes #98 #123 #121 #102  (yay!)

### DIFF
--- a/sample/ManualTestApp/Program.cs
+++ b/sample/ManualTestApp/Program.cs
@@ -32,8 +32,8 @@ namespace ManualTestApp
             var cacheHelper = await CreateCacheHelperAsync().ConfigureAwait(false);
             cacheHelper.RegisterCache(pca.UserTokenCache);
 
-            // The token cache helper provides a high level event that informs apps about added / removed accounts.
-            cacheHelper.CacheChanged += (s, e) =>
+            // Advanced scenario for when 2 or more apps share the same cache             
+            cacheHelper.CacheChanged += (s, e) => // this event is very expensive perf wise
             {
                 Console.BackgroundColor = ConsoleColor.DarkCyan;
                 Console.WriteLine($"Cache Changed, Added: {e.AccountsAdded.Count()} Removed: {e.AccountsRemoved.Count()}");
@@ -276,8 +276,7 @@ namespace ManualTestApp
             {
                 storageProperties = new StorageCreationPropertiesBuilder(
                     Config.CacheFileName,
-                    Config.CacheDir,
-                    Config.ClientId)
+                    Config.CacheDir)
                 .WithLinuxKeyring(
                     Config.LinuxKeyRingSchema,
                     Config.LinuxKeyRingCollection,
@@ -287,6 +286,9 @@ namespace ManualTestApp
                 .WithMacKeyChain(
                     Config.KeyChainServiceName,
                     Config.KeyChainAccountName)
+                .WithCacheChangedEvent( // do NOT use unless really necessary, high perf penalty!
+                    Config.ClientId,
+                    Config.Authority)
                 .Build();
 
                 var cacheHelper = await MsalCacheHelper.CreateAsync(
@@ -305,8 +307,7 @@ namespace ManualTestApp
                 storageProperties =
                     new StorageCreationPropertiesBuilder(
                         Config.CacheFileName,
-                        Config.CacheDir,
-                        Config.ClientId)
+                        Config.CacheDir)
                     .WithLinuxUnprotectedFile()
                     .WithMacKeyChain(
                         Config.KeyChainServiceName,
@@ -318,7 +319,6 @@ namespace ManualTestApp
 
                 return cacheHelper;
             }
-
         }
 
         private static IPublicClientApplication CreatePublicClient(string authority)
--- a/src/Shared/StorageCreationPropertiesBuilder.cs
+++ b/src/Shared/StorageCreationPropertiesBuilder.cs
@@ -12,14 +12,15 @@ namespace Microsoft.Identity.Client.Extensions.Msal
 namespace Microsoft.Identity.Client.Extensions.Web
 #endif
 {
-/// <summary>
-/// An incremental builder for <see cref="StorageCreationProperties"/> objects.
-/// </summary>
-public class StorageCreationPropertiesBuilder
+    /// <summary>
+    /// An incremental builder for <see cref="StorageCreationProperties"/> objects.
+    /// </summary>
+    public class StorageCreationPropertiesBuilder
     {
         private readonly string _cacheFileName;
         private readonly string _cacheDirectory;
-        private readonly string _clientId;
+        private string _clientId;
+        private string _authority;
         private string _macKeyChainServiceName;
         private string _macKeyChainAccountName;
         private string _keyringSchemaName;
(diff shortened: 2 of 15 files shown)


sample/ManualTestApp/Program.cs
```diff
@@ -276,8 +276,7 @@ private static async Task<MsalCacheHelper> CreateCacheHelperAsync()
             {
                 storageProperties = new StorageCreationPropertiesBuilder(
```
bgavrilMS, line 277: Example usage is here - you can see that Client_ID is no longer a mandatory param. If people need CacheChanged event, they need to configure both "client_id" and "authority". If they don't, they'll get a good error message when subscribing to `CacheChanged`

sample/ManualTestApp/Program.cs
```diff
@@ -287,6 +286,9 @@ private static async Task<MsalCacheHelper> CreateCacheHelperAsync()
                 .WithMacKeyChain(
                     Config.KeyChainServiceName,
                     Config.KeyChainAccountName)
+                .WithCacheChangedEvent( // do NOT use unless really necessary, high perf penalty!
```
bgavrilMS, line 289: CacheChanged event is scoped to this (client_id, authority) pair. There isn't (currently) a way to get events about other pairs, except by creating different instances of the MsalCacheHelper object.

src/Shared/StorageCreationPropertiesBuilder.cs
```diff
@@ -37,11 +38,25 @@ public class StorageCreationPropertiesBuilder
         /// <param name="cacheFileName">The name of the cache file to use when creating or opening storage.</param>
         /// <param name="cacheDirectory">The name of the directory containing the cache file.</param>
         /// <param name="clientId">The client id for the calling application</param>
+        [Obsolete("Use StorageCreationPropertiesBuilder(string, string) instead. " +
+            "If you need to consume the CacheChanged event then also use WithCacheChangedEvent(string, string)", false)]
         public StorageCreationPropertiesBuilder(string cacheFileName, string cacheDirectory, string clientId)
```
bgavrilMS, line 43: This constructor is now obsolete.

bgavrilMS:
Also adding @jmprieur for the public API changes. 
Note that I'll be merging this tomorrow.

sample/ManualTestApp/Program.cs
jmprieur: Does it mean that you won't have changes in the FOCI ?

jmprieur (commented):
LGTM.
I don't fully understand how this would be a shared cache from the point of view of the change events?

bgavrilMS:
@jmprieur - for FOCI apps, the client_id is not used in the internal logic of `GetAccounts`. Great question by the way. So no, there won't be any changes to VS or other FOCI apps.

Related to this being a shared cache. Think of this object as a singleton, as it represents a file on disk. I can then register a bunch of `ITokenCache` objects (from different `PublicClientApplication` objs) to synchronize with this file. If I'm interested in knowing which accounts were added / removed from the cache, I can register for an event: 

```csharp 
StoregePropertiesBuilder().WithCacheChangedEvent(client_id, authority);
```
and later subscribe to the event

```csharp
MsalCacheHelper helper = MsalCacheHelper.CreateAsync(storageProperties);
helper.CacheChangedEvent += <listner>
```

The `CacheChangeEvent` will only fire for the (client_id, authority) that you have configured. If you don't configure any, I'll throw.

You cannot currently listen to events for different pairs of (client_id, authority). VS doesn't care about it and nobody else expressed an interest in this event. As such, I think this is the simplest solution. 

A better solution would be to have an API like

```csharp
helper.RegisterCacheChangedCallback(client_id, authority, Action<accounts_added, accounts_removed>)
```

This is doable, but more work. Do you think it's worth doing now?

jmprieur:
Thanks for the explanations, @bgavrilMS 
No, let's not do it until someone expresses their need.

src/Shared/StorageCreationPropertiesBuilder.cs
```diff
@@ -79,6 +95,23 @@ public StorageCreationPropertiesBuilder WithMacKeyChain(string serviceName, stri
...
+        /// Enables the use of the MsalCacheHelper.CacheChanged event, which notifies about
+        /// accounts added and removed. These accounts are scoped to the client_id and authority
+        /// specified here.
+        /// </summary>
+        /// <param name="clientId">The client id for which you wish to receive notifications</param>
+        /// <param name="authority">The authority for which you wish to receive notifications</param>
+        /// <returns>The augmented builder</returns>        
+        public StorageCreationPropertiesBuilder WithCacheChangedEvent(
+            string clientId,
+            string authority = "https://login.microsoftonline.com/common")
```
henrik-me, line 108: >"https://login.microsoftonline.com/common" [](start = 31, length = 42)

this is used in a number of places, should this be a const from MSAL, ensuring we only have one place we need to change default authority?

tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/MsalCacheHelperTests.cs
```diff
@@ -393,7 +419,77 @@ public async Task EventFiresAsync()
             string cacheWithOneUser = File.ReadAllText(
                 ResourceHelper.GetTestResourceRelativePath("token_cache_one_acc_seed.json"));
 
+            var properties = _storageCreationPropertiesBuilder
+                .WithCacheChangedEvent(ClientId, "https://login.microsoftonline.com/common")
```
henrik-me, line 423: >"https://login.microsoftonline.com/common" [](start = 49, length = 42)

any reason this isn't a const like ClientID ?

henrik-me (approved):
:shipit:

PR #131, opened 2021-02-16
Bogavril/yaml release

bgavrilMS:


bgavrilMS:
Need to make some changes to the release build, and it looks like we've never migrated from old style build to yaml build.

bgavrilMS:
Merging so I can continue the work.

PR #133, opened 2021-02-17
Stop keeping references of ITokenCache

bgavrilMS:


src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
```diff
@@ -78,11 +78,7 @@ public class MsalCacheHelper
         /// </remarks>
         public event EventHandler<CacheChangedEventArgs> CacheChanged;
 
-        /// <summary>
-        /// Contains a reference to all caches currently registered to synchronize with this MsalCacheHelper, along with
-        /// timestamp of the cache file the last time they deserialized.
-        /// </summary>
-        internal readonly HashSet<ITokenCache> _registeredCaches = new HashSet<ITokenCache>();
```
bgavrilMS, line 85: This collection is not used for anything, it was supposed to be used with some optimizations that were dropped. Dead code.

tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/MsalCacheHelperTests.cs
```diff
@@ -248,6 +248,31 @@ public void LockTimeoutTest()
...
+
+            // let the top level object go out of scope
+            pca = null;
+
+            // force garbage collection to make sure pca is cleaned (works only in RELEASE mode)
+            GC.Collect();
+            GC.WaitForPendingFinalizers();
+            GC.Collect();
+
+            Assert.IsFalse(weakReference.IsAlive, "The ITokenCache object should have been garbage collected. This indicates a memory leak.");
```
bgavrilMS, line 273: Pretty cool trick to verify that an object is garbage collected. Only works in RELEASE mode though, since in DEBUG mode objects are kept around for longer.

henrik-me (approved):
:shipit:

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
```diff
@@ -255,57 +251,38 @@ public static string UserRootDirectory
...
-
-                tokenCache.SetBeforeAccess(BeforeAccessNotification);
-                tokenCache.SetAfterAccess(AfterAccessNotification);
-
-                _logger.LogInformation($"Initializing msal cache");
-            }
+            _logger.LogInformation($"Registering token cache with on disk storage");
 
-            _registeredCaches.Add(tokenCache); // Ignore return value, since we already bail if _registeredCaches contains tokenCache earlier
+            // If the token cache was already registered, this operation does nothing
```
henrik-me, line 267: >// If the token cache was already registered, this operation does nothing [](start = 12, length = 73)

It depends on the ITokenCache implementation? (Could be a multi cast event handler implementation)

bgavrilMS: We use simple callbacks here, not events. Setting one function pointer would simply replace the other function pointer, if any. See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/a055ebbf6fa08988ce02d1936e636ff1066a0b24/src/client/Microsoft.Identity.Client/TokenCache.Notifications.cs#L87

tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/MsalCacheHelperTests.cs
jennyf19: very cool.

PR #152, opened 2021-09-01
Add WithUnprotectedFile() option

bgavrilMS:


src/Shared/StorageCreationProperties.cs
```diff
@@ -57,6 +59,23 @@ public class StorageCreationProperties
             Authority = authority;
             LockRetryDelay = lockRetryDelay;
             LockRetryCount = lockRetryCount;
+
+            Validate(keyringSecretLabel);
+
+        }
+
+        private void Validate(string keyringSecretLabel)
```
pmaytak, line 67: Does `keyringSecretLabel` need to be passed in or can the `KeyringSecretLabel` property be used?

src/Shared/StorageCreationProperties.cs
```diff
@@ -57,6 +59,23 @@ public class StorageCreationProperties
...
+        {
+            if ((UseLinuxUnencryptedFallback || UseUnencryptedFallback) &&
```
pmaytak, line 69: Is it valid to have plaintext storage on linux but protected on mac or windows?
Is `UseLinuxUnencryptedFallback` then essentially obsolete and superseded by `UseUnencryptedFallback`?

This shows like you can have mixed modes?
https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/0b28722578e7b7263f26fad0cccdf9f0ec48c84b/src/Microsoft.Identity.Client.Extensions.Msal/Storage.cs#L74-L105

pmaytak (changes requested):
Questions about protected/unprotected modes mutual exclusivity,

src/Shared/StorageCreationProperties.cs
bgavrilMS: Fair point, I will make them mutually exclusive. 
Yes, UseUnencryptedFallback is a superset. There are some very rare cases where encryption does not work on Win and Mac. Only PowerShell seems to bump into them. 

So I'd rather not advertise this new option, and let folks use UseLinuxUnencryptedFallback  instead.

pmaytak (approved):

src/Shared/StorageCreationProperties.cs
```diff
@@ -57,6 +59,33 @@ public class StorageCreationProperties
...
+                    !string.IsNullOrEmpty(KeyringCollection)                    ))
+            {
+                throw new ArgumentException("Using plaintext storage is mutually exclusive with other Linux storage options. ");
+            }
+
+            if ((UseUnencryptedFallback ) && 
+                (   !string.IsNullOrEmpty(MacKeyChainServiceName) ||
+                    !string.IsNullOrEmpty(MacKeyChainAccountName)))
+            {
+                throw new ArgumentException("Using Luiplaintext storage is mutually exclusive with other Mac storage options. ");
```
pmaytak, line 86: typo.

PR #165, opened 2022-02-25
Update MSAL version

neha-bhargava:


gladjohn (approved):
looks good @neha-bhargava

PR #167, opened 2022-03-22
Upgrade deprecated TFMs

bgavrilMS:


SameerK-MSFT (approved):

trwalke (approved):

pmaytak (commented):
Why is this change needed, just cleaning up or other reason?
Any risk of breaking users? I assume for net45 and netcore2.1 apps, the Net Standard DLL will get used instead.

bgavrilMS:
> Why is this change needed, just cleaning up or other reason? Any risk of breaking users? I assume for net45 and netcore2.1 apps, the Net Standard DLL will get used instead.

This is to just eliminate some of the security warnings from the build. Also, I can't find the net45 SDK (only net452 is around). 
I am also curious to see if anybody will complain. As you said, netstandard2 will take care of netcore2.1

PR #168, opened 2022-03-22
Set 600 permissions on fallback files

bgavrilMS:
Done: 
- [x] Product code
- [x] Unit tests on Win
- [x] Automation on Unix
- [x] Manual testing on Windows
- [x] Manual testing on Unix

### DIFF
--- a/src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileAccessor.cs
+++ b/src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileAccessor.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.IO;
+using System.Security.AccessControl;
 using System.Text;
 
 namespace Microsoft.Identity.Client.Extensions.Msal
@@ -13,10 +14,12 @@ namespace Microsoft.Identity.Client.Extensions.Msal
 
         private readonly string _cacheFilePath;
         private readonly TraceSourceLogger _logger;
+        private readonly bool _useChmod;
 
-        public FileAccessor(string cacheFilePath, TraceSourceLogger logger)
+        public FileAccessor(string cacheFilePath, bool useChmod, TraceSourceLogger logger)
         {
             _cacheFilePath = cacheFilePath;
+            _useChmod = useChmod;
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
         }
 
@@ -28,7 +31,7 @@ namespace Microsoft.Identity.Client.Extensions.Msal
 
         public ICacheAccessor CreateForPersistenceValidation()
         {
-            return new FileAccessor(_cacheFilePath + ".test", _logger);
+            return new FileAccessor(_cacheFilePath + ".test", _useChmod, _logger);
         }
 
         public byte[] Read()
@@ -53,7 +56,7 @@ namespace Microsoft.Identity.Client.Extensions.Msal
 
         public void Write(byte[] data)
         {
-            FileIOWithRetries.WriteDataToFile(_cacheFilePath, data, _logger);
+            FileIOWithRetries.CreateAndWriteToFile(_cacheFilePath, data, true, _logger);
         }
     }
 }
--- a/src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileAccessor.cs
+++ b/src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileAccessor.cs
@@ -14,12 +14,12 @@ namespace Microsoft.Identity.Client.Extensions.Msal
 
         private readonly string _cacheFilePath;
         private readonly TraceSourceLogger _logger;
-        private readonly bool _useChmod;
+        private readonly bool _setOwnerOnlyPermission;
 
-        public FileAccessor(string cacheFilePath, bool useChmod, TraceSourceLogger logger)
+        internal FileAccessor(string cacheFilePath, bool setOwnerOnlyPermissions, TraceSourceLogger logger)
         {
             _cacheFilePath = cacheFilePath;
-            _useChmod = useChmod;
+            _setOwnerOnlyPermission = setOwnerOnlyPermissions;
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
         }
 
@@ -31,7 +31,7 @@ namespace Microsoft.Identity.Client.Extensions.Msal
 
         public ICacheAccessor CreateForPersistenceValidation()
         {
-            return new FileAccessor(_cacheFilePath + ".test", _useChmod, _logger);
+            return new FileAccessor(_cacheFilePath + ".test", _setOwnerOnlyPermission, _logger);
         }
 
         public byte[] Read()
@@ -56,7 +56,7 @@ namespace Microsoft.Identity.Client.Extensions.Msal
 
         public void Write(byte[] data)
         {
-            FileIOWithRetries.CreateAndWriteToFile(_cacheFilePath, data, true, _logger);
+            FileIOWithRetries.CreateAndWriteToFile(_cacheFilePath, data, _setOwnerOnlyPermission, _logger);
         }
     }
 }
--- a/src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs
+++ b/src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs
@@ -1,0 +1,136 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Security.AccessControl;
+using System.Security.Principal;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.Extensions.Msal.Accessors
+{
+    internal static class FileWithPermissions
+    {
+        [DllImport("libc", SetLastError = true)]
+        private static extern int chmod(string pathname, int mode);
+
+        // user permissions
+        const int S_IRUSR = 0x100;
+        const int S_IWUSR = 0x80;
+        const int S_IXUSR = 0x40;
+
+        // group permission
+        const int S_IRGRP = 0x20;
+        const int S_IWGRP = 0x10;
+        const int S_IXGRP = 0x8;
+
+        // other permissions
+        const int S_IROTH = 0x4;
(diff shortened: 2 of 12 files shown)


src/Microsoft.Identity.Client.Extensions.Msal/Microsoft.Identity.Client.Extensions.Msal.csproj
```diff
@@ -45,6 +45,7 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetCore)' or '$(TargetFramework)' == '$(TargetFrameworkNetStandard)'">
     <PackageReference Include="System.Security.Cryptography.ProtectedData" Version="4.5.0" />
+    <PackageReference Include="System.IO.FileSystem.AccessControl" Version="5.0.0" />
```
bgavrilMS, line 48: NET core does not have the file permissions stuff by default, it's in a separate package. This logic is Windows only. For Unix, we need to rely on `chmod`

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileAccessor.cs
```diff
@@ -13,10 +14,12 @@ internal class FileAccessor : ICacheAccessor
 
         private readonly string _cacheFilePath;
         private readonly TraceSourceLogger _logger;
+        private readonly bool _useChmod;
 
-        public FileAccessor(string cacheFilePath, TraceSourceLogger logger)
+        public FileAccessor(string cacheFilePath, bool useChmod, TraceSourceLogger logger)
```
SameerK-MSFT, line 19: I was initially concerned about having non-backward compatible change in public constructor. But realized that the class itself is internal. Can the constructor and other public methods be internal too?

SameerK-MSFT (approved):

trwalke (approved):

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs
```diff
@@ -0,0 +1,136 @@
...
+                throw new PlatformNotSupportedException();
+            }
+        }
+
+        /// <summary>
+        /// Based on https://stackoverflow.com/questions/45132081/file-permissions-on-linux-unix-with-net-core
+        /// </summary>
+        private static void WriteToNewFileWith600PermissionsUnix(string filePath, byte[] data)
+        {
+            File.WriteAllBytes(filePath, data);
```
pmaytak, line 60: So we have to create a file first, then set permissions on it. If the set fails, we try to delete the file. But there's a case if the delete fails, then the unprotected file still remains on a disk? Although perhaps an improbable scenario.

src/Microsoft.Identity.Client.Extensions.Msal/FileIOWithRetries.cs
```diff
@@ -33,15 +34,24 @@ internal static void DeleteCacheFile(string filePath, TraceSourceLogger logger)
             }, logger);
         }
 
-        internal static void WriteDataToFile(string filePath, byte[] data, TraceSourceLogger logger)
+        internal static void CreateAndWriteToFile(string filePath, byte[] data, bool setChmod600, TraceSourceLogger logger)
```
pmaytak, line 37: nit: -> `bool setOwnerOnlyReadWrite`, more descriptive, it's cross-platform method.

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileAccessor.cs
```diff
@@ -13,10 +14,12 @@ internal class FileAccessor : ICacheAccessor
 
         private readonly string _cacheFilePath;
         private readonly TraceSourceLogger _logger;
+        private readonly bool _useChmod;
```
pmaytak, line 17: Similar naming comment as below.

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileAccessor.cs
```diff
@@ -53,7 +56,7 @@ public byte[] Read()
 
         public void Write(byte[] data)
         {
-            FileIOWithRetries.WriteDataToFile(_cacheFilePath, data, _logger);
+            FileIOWithRetries.CreateAndWriteToFile(_cacheFilePath, data, true, _logger);
```
pmaytak, line 59: Why `_useChmod` not used? In `DpApiEncryptedFileAccessor` `false` is passed to `FileAccessor`, but here it's true.

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        /// <summary>
+        /// Windows has a more complex file security system. "600" mode, i.e. read/write for owner translates to this in Windows.
+        /// </summary>
+        /// <param name="filePath"></param>
+        /// <param name="data"></param>
+        private static void WriteToNewFileWith600PermissionsWindows(string filePath, byte[] data)
+        {
+            FileSecurity security = new FileSecurity();
```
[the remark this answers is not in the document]
bgavrilMS: I don't know of any, I don't think so. You should always have permissions to limit the R / W access to a file that you own.

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// <summary>
+        /// Based on https://stackoverflow.com/questions/45132081/file-permissions-on-linux-unix-with-net-core
+        /// </summary>
+        private static void WriteToNewFileWithOwnerRWPermissionsUnix(string filePath, byte[] data)
+        {
+            File.Create(filePath);
+
+            // Setting permissions to 0600 
+            const int _0600 = S_IRUSR | S_IWUSR; // read & write only for user
+            int result = chmod(filePath, _0600);
```
KalleOlaviNiemitalo, line 69: This seems to have a race condition still — if File.Create uses looser permissions and a process of another user opens the file for reading between File.Create and chmod, then that process can keep reading subsequent changes of the file via that open file description, even after chmod changes the access rights such that the other user is no longer able to open the file.

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs
```diff
@@ -0,0 +1,137 @@
...
+                try
+                {
+                    File.Delete(filePath);
+                }
+                catch (Exception)
+                {
+                    // ignored
+                }
+
+                throw new IOException($"Failed to set permissions on file {filePath} - error code {result}");
```
KalleOlaviNiemitalo, line 82: The chmod function returns 0 on success or -1 on error. This then always logs -1 as the error code, if chmod failed. If you want to log the actual error code, this needs to call Marshal.GetLastWin32Error between chmod and File.Delete.

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs
```diff
@@ -0,0 +1,137 @@
...
+                throw new PlatformNotSupportedException();
+            }
+        }
+
+        /// <summary>
+        /// Based on https://stackoverflow.com/questions/45132081/file-permissions-on-linux-unix-with-net-core
+        /// </summary>
+        private static void WriteToNewFileWithOwnerRWPermissionsUnix(string filePath, byte[] data)
+        {
+            File.Create(filePath);
```
KalleOlaviNiemitalo, line 65: File.Create here returns a FileStream that should preferably be closed explicitly, instead of leaving that to the finaliser.

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs
KalleOlaviNiemitalo: Perhaps you can reuse the code from <https://github.com/NuGet/NuGet.Client/commit/d62db666c710bf95121fe8f5c6a6cbe01985456f>.

bgavrilMS: There is a lock mechanism (also uses files) in place which protects against another process making changes to the same file.

KalleOlaviNiemitalo: From <https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3186>, I got the impression that this chmod was intended for security against malicious users, who would not use the lock mechanism if they want to read the files of another user.

bgavrilMS: It's a good point, although I do not see an attack pattern in re-writing the user's token cache, as those tokens are all digitally signed. Stealing them is the problem we are trying to solve. 

Can you please log a bug for this? Or maybe you are interested in porting the nuget code directly, i.e. create a PR? I will help with testing.

KalleOlaviNiemitalo: I filed the issue <https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/issues/174> for this.

KalleOlaviNiemitalo: I filed the issue <https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/issues/174> for this.

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs
KalleOlaviNiemitalo: You added a `Marshal.GetLastWin32Error` call in <https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/pull/173>, but it's done _after_  the `File.Delete` call, so if `File.Delete` changes the last error, then the `IOException` won't indicate the error code that caused `chmod` to fail.

PR #172, opened 2022-04-07
Test Fix

bgavrilMS:


pmaytak:
I made this change already in an earlier PR, but [Linux and Mac tests are still failing](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=905419&view=results), seems like something else is going on.

PR #173, opened 2022-04-10
IO fix

bgavrilMS:


bgavrilMS:
PR build: https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=906758&view=results

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs
```diff
@@ -62,14 +64,16 @@ public static void WriteToNewFileWithOwnerRWPermissions(string path, byte[] data
...
+            {
+            }
 
             // Setting permissions to 0600 
             const int _0600 = S_IRUSR | S_IWUSR; // read & write only for user
             int result = chmod(filePath, _0600);
 
             if (result != 0)
-            {
+            {
```
jmprieur, line 76: extra space?

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs
```diff
@@ -62,14 +64,16 @@ public static void WriteToNewFileWithOwnerRWPermissions(string path, byte[] data
         /// </summary>
         private static void WriteToNewFileWithOwnerRWPermissionsUnix(string filePath, byte[] data)
         {
-            File.Create(filePath);
+            using (File.Create(filePath))
+            {
```
jmprieur, line 68: Why is it? Is it to be sure the file is indeed created?
Do you need to create the folder before?

jmprieur (approved):
LGTM
Thanks @bgavrilMS

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs
bgavrilMS: File.Create returns a stream for writing.

pmaytak (commented):
Any idea why this fixed the Mac and Linux tests?

bgavrilMS:
> Any idea why this fixed the Mac and Linux tests?

Yes, because `File.Open` actually returns a FileStream, i.e. keeps the file open. Looks like on Windows the GC closes the stream quickly, but not on Mac / Linux. After applying the permissions, I re-open the file to write to it (in Append mode)

Issue #174, opened 2022-04-12
Race condition in setting 600 permissions


KalleOlaviNiemitalo:
[FileWithPermissions.WriteToNewFileWithOwnerRWPermissionsUnix](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/93354933e6676c86a7d9cb479fdfe8f21aecf82c/src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs#L62-L93) first calls `File.Create` to create the file with default permissions, and then calls `chmod` to change the permissions to 600 (octal) so that the owner has read and write access, and the group and other users have no access. This `chmod` is intended to prevent malicious users from reading secrets from the file, even if the `umask` of the process allows read access to be granted to such users.

However, if a malicious user is able to open the file between the `File.Create` and `chmod` calls, then the malicious user can keep the file descriptor open, wait for the library to write secrets to the file, and use the file descriptor to read the file. The operating system will allow this read even though `chmod` already changed the access rights so that the malicious user is no longer able to open the file. Although the library has a locking mechanism to prevent simultaneous access to the file, malicious processes cannot be assumed to use the locking mechanism.

To prevent this, the library should atomically set the 600 access mode when it creates the file, e.g. via the `mode_t mode` parameter of [`open` on Linux](<https://man7.org/linux/man-pages/man2/open.2.html> "open(2) - Linux manual page"). There is no API for that in .NET 6, but one is being proposed in <https://github.com/dotnet/runtime/issues/67837>. (In the C standard library, [`fopen_s`](<https://en.cppreference.com/w/c/io/fopen> "fopen, fopen_s - cppreference.com") sets a restrictive access mode, unlike the older `fopen` function. However, the resulting `FILE *` would not be trivial to use in .NET, and AFAIK the GNU C Library does not implement `fopen_s`.)

Alternatively, if a `revoke(const char *path)` or `frevoke(int fd)` system call were available, then the library could use one of those to revoke the file descriptors of any malicious processes after `chmod`.

_Originally reported in <https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/pull/168#discussion_r843430829>._

pmaytak:
@KalleOlaviNiemitalo Thanks for the informative feedack.

PR #175, opened 2022-04-12
Fix for #174 - on *nix, create file and apply permissions at the same…

bgavrilMS:
… time

### DIFF
--- a/src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs
+++ b/src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs
@@ -18,23 +18,17 @@ namespace Microsoft.Identity.Client.Extensions.Msal.Accessors
 
         // See https://ss64.com/bash/chmod.html 
 
-        [DllImport("libc", SetLastError = true)]
-        private static extern int chmod(string pathname, int mode);
 
-        // user permissions
-        const int S_IRUSR = 0x100;
-        const int S_IWUSR = 0x80;
-        const int S_IXUSR = 0x40;
+        [DllImport("libc", EntryPoint = "creat", SetLastError = true)]
+        private static extern int PosixCreate([MarshalAs(UnmanagedType.LPStr)] string pathname, int mode);
 
-        // group permission
-        const int S_IRGRP = 0x20;
-        const int S_IWGRP = 0x10;
-        const int S_IXGRP = 0x8;
+        [DllImport("libc", EntryPoint = "chmod", SetLastError = true)]
+        private static extern int PosixChmod([MarshalAs(UnmanagedType.LPStr)] string pathname, int mode);
 
-        // other permissions
-        const int S_IROTH = 0x4;
-        const int S_IWOTH = 0x2;
-        const int S_IXOTH = 0x1;
+        [DllImport("libc", EntryPoint = "umask", SetLastError = true)]
+        private static extern int PosixUMask(int mask);
+
+
         #endregion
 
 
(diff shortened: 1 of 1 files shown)


bgavrilMS:
Not sure why the CI build doesn't trigger, but a manual trigger shows this works: https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=907401&view=results

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs
```diff
@@ -60,35 +58,40 @@ public static void WriteToNewFileWithOwnerRWPermissions(string path, byte[] data
...
+        private static void WriteToNewFileWithOwnerRWPermissionsUnix(string path, byte[] data)
         {
-            using (File.Create(filePath))
+            int _0600 = Convert.ToInt32("600", 8);
+           
+            int fd;
+
+            try
             {
+                fd = PosixCreate(path, _0600);
```
pmaytak, line 72: Do I understand correctly that `creat()` function overwrites an existing file? Maybe just add to a method comment that existing files are overwritten.

>    [creat()](https://man7.org/linux/man-pages/man2/open.2.html)
>        A call to creat() is equivalent to calling open() with flags
>        equal to O_CREAT|O_WRONLY|O_TRUNC.

>        O_TRUNC
>               If the file already exists and is a regular file and the
>               access mode allows writing (i.e., is O_RDWR or O_WRONLY)
>               it will be truncated to length 0.  If the file is a FIFO
>               or terminal device file, the O_TRUNC flag is ignored.
>               Otherwise, the effect of O_TRUNC is unspecified.

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs
```diff
@@ -60,35 +58,40 @@ public static void WriteToNewFileWithOwnerRWPermissions(string path, byte[] data
...
+            }
+            catch (Exception exception)
```
pmaytak, line 74: `creat` doesn't throw right? This would only catch interop related errors? So we would need to log the last Win32 error in the next `if` statement if `fd == -1`.

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs
```diff
@@ -60,35 +58,40 @@ public static void WriteToNewFileWithOwnerRWPermissions(string path, byte[] data
...
+            {
+                using (File.Create(path))
                 {
-                    // ignored
+                    // File.Create() should have thrown an exception with an appropriate error message
                 }
-
-                throw new IOException($"Failed to set permissions on file {filePath} Error code: {result} Last error: {Marshal.GetLastWin32Error()}");
+                File.Delete(path);
+                throw new InvalidOperationException("libc creat failed, but File.Create did not");
```
pmaytak, line 88: Why do we need to do this block of code? Maybe add a comment + descriptive error message.

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs
```diff
@@ -60,35 +58,40 @@ public static void WriteToNewFileWithOwnerRWPermissions(string path, byte[] data
...
             }
 
-            using (var stream = new FileStream(filePath, FileMode.Append))
+            var sfh = new SafeFileHandle((IntPtr)fd, ownsHandle: true);
+            using (var fs = new FileStream(sfh, FileAccess.ReadWrite))
```
pmaytak, line 92: `safeFileHandle`, `fileStream`

pmaytak (approved):
I think just need to refactor print out correct error + comments + nits.

src/Microsoft.Identity.Client.Extensions.Msal/Accessors/FileWithPermissions.cs
bgavrilMS: Code was copied from nuget repo ... But I'll add a comment.

PR #178, opened 2022-05-18
Optimized for perf

SameerK-MSFT:
GetProcess is called only once and formatting is done only once

### DIFF
--- a/src/Shared/SharedUtilities.cs
+++ b/src/Shared/SharedUtilities.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Threading;
@@ -30,6 +31,8 @@ namespace Microsoft.Identity.Client.Extensions.Web
         private static readonly string s_usernameEnvVar = Environment.GetEnvironmentVariable("USERNAME");
 
         private static readonly Lazy<bool> s_isMono = new Lazy<bool>(() => Type.GetType("Mono.Runtime") != null);
+
+        private static readonly Lazy<Process> s_currentProcess = new Lazy<Process>(() => Process.GetCurrentProcess());
 
         /// <summary>
         ///  Is this a windows platform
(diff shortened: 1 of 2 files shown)


src/Shared/SharedUtilities.cs
```diff
@@ -31,6 +32,8 @@ public static class SharedUtilities
 
         private static readonly Lazy<bool> s_isMono = new Lazy<bool>(() => Type.GetType("Mono.Runtime") != null);
 
+        private static readonly Lazy<Process> s_currentProcess = new Lazy<Process>(() => Process.GetCurrentProcess());
```
pmaytak, line 35: Hmm, wondering if there are could be issues to staticly hanging onto a process reference? Recommendation [here](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process?view=net-6.0#remarks) is to dispose of Process after being used.

Maybe save a string instead something like this?
```suggestion
        private static readonly Lazy<string> s_currentProcessInfo = new Lazy<string>(() => $"{Process.GetCurrentProcess().Id} {Process.GetCurrentProcess().ProcessName}");
```

src/Shared/SharedUtilities.cs
```diff
@@ -77,6 +80,18 @@ internal static bool IsMonoPlatform()
             return s_isMono.Value;
         }
 
+        /// <summary>
+        /// Instantiates the process if not done already and retrieves the ID the process.
+        /// </summary>
+        /// <returns>process id</returns>
+        public static int GetCurrentProcessId() => s_currentProcess.Value.Id;
```
pmaytak, line 87: internal?

pmaytak (changes requested):

src/Shared/SharedUtilities.cs
SameerK-MSFT: My thought behind was that this is a static variable. So, the dispose would not be necessary. And keeping it there would help in accessing more properties in the future. I am open either way.

src/Shared/SharedUtilities.cs
SameerK-MSFT: This being a utility, public would be good. Anyone can use it.

src/Shared/SharedUtilities.cs
bgavrilMS: We want to cache a string, not the Process. Seems less risky with what Peter is proposing.

SameerK-MSFT: I think we can cache id and name separately and create string one time before the loop. So Id and process can be with no context in the shared utility and used later by anyone. Creating formatted string before the loop should not be costly. Also, the format can be changed anytime.

bgavrilMS (approved):

gladjohn (approved):
looks good @SameerK-MSFT

pmaytak (approved):
I'm a bit reserved in this perspective - the only change I would make the Get methods internal, and expose them only if there's a specific need by users.

SameerK-MSFT:
Shows that the pipeline was successful. https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=926323&view=results

PR #185, opened 2022-08-01
Ref updates

bgavrilMS:
Eliminates a use of newtonsoft 9


bgavrilMS:
https://identitydivision.visualstudio.com/IDDP/_build?definitionId=1071

pmaytak (approved):
Weird that the PR builds don't trigger. YAML looks write and point to main branch...

PR #190, opened 2022-08-26
Integrating with MSAL's IIdentityLogger

trwalke:
fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3627

### DIFF
--- a/src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
+++ b/src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
@@ -7,6 +7,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
+using Microsoft.IdentityModel.Abstractions;
 
 namespace Microsoft.Identity.Client.Extensions.Msal
 {
@@ -139,8 +140,8 @@ namespace Microsoft.Identity.Client.Extensions.Msal
                     }
                     catch (Exception e)
                     {
-                        logger.LogError("An error occured while reading the token cache: " + e);
-                        logger.LogError("Deleting the token cache as it might be corrupt.");
+                        LogMessage(args.IdentityLogger, EventLogLevel.Error, "[Microsoft.Identity.Client.Extensions] An error occured while reading the token cache: " + e, logger);
+                        LogMessage(args.IdentityLogger, EventLogLevel.Error, "[Microsoft.Identity.Client.Extensions] Deleting the token cache as it might be corrupt.", logger);
                         tempCache.Clear(ignoreExceptions: true);
                     }
                 });
@@ -397,16 +398,17 @@ namespace Microsoft.Identity.Client.Extensions.Msal
         /// <param name="args">Callback parameters from MSAL</param>
         internal void BeforeAccessNotification(TokenCacheNotificationArgs args)
         {
-            _logger.LogInformation($"Before access");
+            LogMessage(args.IdentityLogger, EventLogLevel.Informational, $"Before access");
 
-            _logger.LogInformation($"Acquiring lock for token cache");
+            LogMessage(args.IdentityLogger, EventLogLevel.Informational, $"Acquiring lock for token cache");
 
             // OK, we have two nested locks here. We need to maintain a clear ordering to avoid deadlocks.
             // 1. Use the CrossPlatLock which is respected by all processes and is used around all cache accesses.
             // 2. Use _lockObject which is used in UnregisterCache, and is needed for all accesses of _registeredCaches.
             CacheLock = CreateCrossPlatLock(_storageCreationProperties);
 
-            _logger.LogInformation($"Before access, the store has changed");
+            LogMessage(args.IdentityLogger, EventLogLevel.Informational, $"Before access, the store has changed");
+
             byte[] cachedStoreData = null;
             try
             {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
```diff
@@ -397,40 +398,41 @@ private static CrossPlatLock CreateCrossPlatLock(StorageCreationProperties stora
         /// <param name="args">Callback parameters from MSAL</param>
         internal void BeforeAccessNotification(TokenCacheNotificationArgs args)
         {
-            _logger.LogInformation($"Before access");
+            LogMessage(args.IdentityLogger, EventLogLevel.Informational, $"Before access");
```
bgavrilMS, line 401: Should really be "verbose" level

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
```diff
@@ -506,5 +509,33 @@ public void VerifyPersistence()
...
         }
+
+        //Logs to TraceSourceLogger and Identity Logger acquired from MSAL's TokenCacheNotificationArgs
+        private void LogMessage(IIdentityLogger identityLogger, EventLogLevel level, string message)
+        {
+            LogMessage(identityLogger, level, message, _logger);
+        }
+
+        //Logs to TraceSourceLogger and Identity Logger acquired from MSAL's TokenCacheNotificationArgs
+        private static void LogMessage(IIdentityLogger identityLogger, EventLogLevel level, string message, TraceSourceLogger traceSourceLogger)
```
bgavrilMS, line 520: I don't see the point in having this method. It only seems used from the method at line 514.

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
```diff
@@ -139,8 +140,8 @@ public class MsalCacheHelper
                     }
                     catch (Exception e)
                     {
-                        logger.LogError("An error occured while reading the token cache: " + e);
-                        logger.LogError("Deleting the token cache as it might be corrupt.");
+                        LogMessage(args.IdentityLogger, EventLogLevel.Error, "[Microsoft.Identity.Client.Extensions] An error occured while reading the token cache: " + e, logger);
```
bgavrilMS, line 143: Having that prefix there is great - can you add it to all messages please? `[Microsoft.Identity.Client.Extensions]`

bgavrilMS (changes requested):
All Info level msg should really be Verbose
Log actual exceptions or propagate logger (but that would be hard)

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
trwalke: The method at line 514 is a static method so there is no object reference to use when logging. That is why I added one method for static and an overload that is not.

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
trwalke: ok, changing them to verbose. I'll leave the error level though

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
bgavrilMS: Ok, I don't see the point in the method at line 518 then :). What's the point of a static if it needs to use an instance member? 

(this is a non-blocking comment)

bgavrilMS (approved):

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
```diff
@@ -139,8 +138,8 @@ public class MsalCacheHelper
                     }
                     catch (Exception e)
                     {
-                        logger.LogError("An error occured while reading the token cache: " + e);
-                        logger.LogError("Deleting the token cache as it might be corrupt.");
+                        LogMessage(args.IdentityLogger, EventLogLevel.Error, "An error occured while reading the token cache: " + e, logger);
+                        LogMessage(args.IdentityLogger, EventLogLevel.Error, "[Microsoft.Identity.Client.Extensions] Deleting the token cache as it might be corrupt.", logger);
```
pmaytak, line 142: the prefix is not needed, since it's added in LogMessage method.

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
```diff
@@ -397,40 +396,41 @@ private static CrossPlatLock CreateCrossPlatLock(StorageCreationProperties stora
         /// <param name="args">Callback parameters from MSAL</param>
         internal void BeforeAccessNotification(TokenCacheNotificationArgs args)
         {
-            _logger.LogInformation($"Before access");
+            LogMessage(args.IdentityLogger, EventLogLevel.Verbose, $"Before access");
 
-            _logger.LogInformation($"Acquiring lock for token cache");
+            LogMessage(args.IdentityLogger, EventLogLevel.Verbose, $"Acquiring lock for token cache");
```
pmaytak, line 401: I would maybe combine the consecutive log lines with the same log level to improve perf (maybe add NewLine if needed too).

pmaytak (approved):

src/Microsoft.Identity.Client.Extensions.Msal/MsalCacheHelper.cs
trwalke: So both methods arent 100% needed. Its just the method one line 514 reduces the need to pass in one additional variable which makes the code look a little bit cleaner.

PR #191, opened 2022-09-16
Update Msal version to 4.47

neha-bhargava:


gladjohn (approved):
Thanks @neha-bhargava, looks good

